PR #11, opened 2020-04-06
Ddp 3730 date pipe

SimoneMaiwald:


client/src/app/filter-column/models/column.model.ts
```diff
@@ -34,71 +34,73 @@ export class ParticipantColumn {
...
   public static MR_DOCUMENT = new ParticipantColumn("MR Document", "mrDocument", "m");
-  public static MR_DOCUMENT_FILES = new ParticipantColumn("MR File Names", "mrDocumentFileNames", "m");
+  public static MR_DOCUMENT_FILES = new ParticipantColumn("MR Document File Names", "mrDocumentFileNames", "m");
   public static MR_PROBLEM = new ParticipantColumn("MR Problem", "mrProblem", "m");
   public static MR_PROBLEM_TEXT = new ParticipantColumn("MR Problem Text", "mrProblemText", "m");
-  public static MR_UNABLE_TO_OBTAIN = new ParticipantColumn("MR Unable to Obtain", "unableObtain", "m");
-  public static MR_DUPLICATE = new ParticipantColumn("MR Duplicate", "duplicate", "m");
-  public static MR_INTERNATIONAL = new ParticipantColumn("MR International", "international", "m");
-  public static MR_PAPER_CR = new ParticipantColumn("MR Paper C/R Required", "crRequired", "m");
+  public static MR_UNABLE_TO_OBTAIN = new ParticipantColumn("Unable to Obtain", "unableObtain", "m");
```
pegahtah: I think it is better to keep the MR in the name so it is easy to know which is which while looking at the table if both unable to obtains are selected

client/src/app/filter-column/models/column.model.ts
```diff
@@ -34,71 +34,73 @@ export class ParticipantColumn {
...
+  public static MR_DUPLICATE = new ParticipantColumn("Duplicate", "duplicate", "m");
+  public static MR_INTERNATIONAL = new ParticipantColumn("International", "international", "m");
+  public static MR_PAPER_CR = new ParticipantColumn("Paper C/R required", "crRequired", "m");
   public static PATHOLOGY_RESENT = new ParticipantColumn("Pathology Present", "pathologyPresent", "m");
-  public static MR_NOTES = new ParticipantColumn("MR Notes", "mrNotes", "m");
+  public static MR_NOTES = new ParticipantColumn("Notes", "mrNotes", "m");
```
pegahtah: Same thing for notes

client/src/app/filter-column/models/column.model.ts
SimoneMaiwald: true, I thought it is clear, by which dropdown shows the column, but you are right the header will not have that information anymore

pegahtah (approved):

PR #33, opened 2020-06-24
Hello, AppEngine, TCell, and Java11

zyme:
These changes enable deployment of the backend to GAE.  Included here are DDP-4754  and DDP-4754.  One thing to get familiar with is GCP's [secret manager](https://cloud.google.com/secret-manager) (SM), which is proving to be way more fun than vault.  Versions!  UI!  Good CLI tools!  I took the opportunity to replace a few vault interactions with SM as part of GAE.

Along the way, expanded logging on the backend to include real IP and user email, as captured from JWT.

EDIT: Also upgraded to Java11.


pom.xml
```diff
@@ -10,13 +10,45 @@
...
+                <exclusion>
+                    <!-- this pom will declare the version of sparkjava -->
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>spark-core</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+
+        <dependency>
```
zyme, line 47: These are necessary for using cloudsql  db connections.

pom.xml
```diff
@@ -10,13 +10,45 @@
...
     <dependencies>
 
+        <dependency>
+            <groupId>com.sparkjava</groupId>
+            <artifactId>spark-core</artifactId>
+            <version>2.8.0</version>
+            <exclusions>
+                <exclusion>
+                    <!-- this pom will declare the version of sparkjava -->
+                    <groupId>org.slf4j</groupId>
```
zyme, line 35: Needed an upgrade here for `JettyConfig` and getting the proper source IP into logs.

pom.xml
```diff
@@ -152,72 +191,20 @@
                 </configuration>
             </plugin>
             <plugin>
-	      <!-- used to generate single jar file for deployment -->
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-shade-plugin</artifactId>
```
zyme: At irregular intervals, `gcloud` was complaining about the size of the jar.  So instead of doing an uberjar, I broke this out into uploads of all dependent jars.

pom.xml
```diff
@@ -152,72 +191,20 @@
...
-                    <execution>
-                        <phase>package</phase>
-                        <goals>
-                            <goal>shade</goal>
-                        </goals>
-                    </execution>
-                </executions>
-            </plugin>
-            <plugin>
-                <groupId>org.eluder.coveralls</groupId>
```
zyme: All this  test coverage stuff wasn't doing anything, so I removed it.

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -668,4 +685,31 @@ public static boolean isTest(@NonNull String shipperName) {
         }
         return true;
     }
+
+
+    // todo  arz fixme read from config file
+    private static void enableCORS(String methods, String headers) {
```
zyme: Although there is  already a `protected` version of this method on the superclass, I wanted to add the ability to support multiple URLs.  Browsers were giving me guff about using `*` and `included` token auth, so I addressed it here by echoing back the appropriate URL dynamically.

src/main/java/org/broadinstitute/dsm/jetty/JettyConfig.java
```diff
@@ -0,0 +1,153 @@
...
+import spark.embeddedserver.EmbeddedServerFactory;
+import spark.embeddedserver.EmbeddedServers;
+import spark.embeddedserver.jetty.EmbeddedJettyServer;
+import spark.embeddedserver.jetty.HttpRequestWrapper;
+import spark.embeddedserver.jetty.JettyServerFactory;
+import spark.http.matching.MatcherFilter;
+import spark.route.Routes;
+import spark.staticfiles.StaticFilesConfiguration;
+
+public class JettyConfig {
```
zyme, line 27: Copy pasted from study-server.

src/main/resources/log4j.xml
```diff
@@ -19,7 +19,6 @@
     <root>
         <priority value ="info" />
         <appender-ref ref="console" />
-        <appender-ref ref="file" />
```
zyme: Logs should go to stdout so that they are picked up by appsec via stackdriver.

appengine/deploy/StudyManager.yaml
```diff
@@ -0,0 +1,15 @@
+service: study-manager-backend
+runtime: java11
+instance_class: B4
+
+manual_scaling:
+  instances: 1
```
yufengwng: If we're doing manual scaling, we should definitely implement GAE's lifecycle call `/_ah/start`, otherwise GAE will continuously reboot the instance. Since `DSMServer` is running a Spark server, it should be returning `404` right now, which GAE's documentation says is fine. But I'll recommend adding a route and returning a `200`.

appengine/deploy/StudyManager.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+network:
+  instance_tag: study-manager
+
+entrypoint: java -cp dsm-backend-SNAPSHOT.jar -javaagent:tcell/tcellagent.jar -Xmx820m -Xmx820m -Dconfig.file=vault.conf org.broadinstitute.dsm.DSMServer
```
yufengwng: Did we mean `-Xms`?

pom.xml
```diff
@@ -10,13 +10,45 @@
...
 
+        <dependency>
+            <groupId>com.sparkjava</groupId>
+            <artifactId>spark-core</artifactId>
+            <version>2.8.0</version>
+            <exclusions>
+                <exclusion>
+                    <!-- this pom will declare the version of sparkjava -->
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>spark-core</artifactId>
```
yufengwng: I'm a bit confused by this `exclusion`...

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -668,4 +685,31 @@ public static boolean isTest(@NonNull String shipperName) {
...
+            String accessControlRequestHeaders = request.headers("Access-Control-Request-Headers");
+            if (accessControlRequestHeaders != null) {
+                response.header("Access-Control-Allow-Headers", accessControlRequestHeaders);
+            }
+
+            String accessControlRequestMethod = request.headers("Access-Control-Request-Method");
+            if (accessControlRequestMethod != null) {
+                response.header("Access-Control-Allow-Methods", accessControlRequestMethod);
+            }
+
```
yufengwng, line 962: I recall @MocanaAtBroad added some magic header to facilitate better caching, something about `Access-Control-Max-Age`. Should we pull that over as well?

appengine/deploy/build-and-deploy.sh
```diff
@@ -0,0 +1,24 @@
...
+
+#  run the build
+echo "Running maven"
+mvn -DskipTests clean install package -f ../../pom.xml
+
+# bundling dependencies
+rm -fr lib
+mkdir -p lib
+mvn -f ../../pom.xml dependency:copy-dependencies -DoutputDirectory=./appengine/deploy/lib
+cp ../../target/dsm-backend-SNAPSHOT.jar .
```
MocanaAtBroad: Consider giving dsm jar a nicer name.

appengine/deploy/build-and-deploy.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+PROJECT_ID=$1
+
+echo "Reading configs from cloud secret manager"
+gcloud --project=${PROJECT_ID} secrets versions access latest --secret="study-manager-config" > vault.conf
```
MocanaAtBroad: vault integration via files? 😸

appengine/deploy/StudyManager.yaml
MocanaAtBroad: Implementing /_ah/start is a good idea. Don't think it is as important for manual scaling as there will be no up/downscaling, but it is not hard to do. Check Yufengs implementation; wait to respond to /_ah/start request until app is completely initialized.
/_ah/stop too if there is cleanup needed when shutting down.

appengine/deploy/StudyManager.yaml
MocanaAtBroad: Also consider defining a entry point for the jar so class name is not needed in command line.

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -54,9 +58,18 @@
 
     private static final Logger logger = LoggerFactory.getLogger(DSMServer.class);
 
+    private static final List<String> ALLOWED_ORIGINS = Arrays.asList("http://localhost:4200","https://dsm.datadonationplatform.org",
+            "https://dsm.dev.datadonationplatform.org","https://dsm.test.datadonationplatform.org",
+            "https://dsm.staging.datadonationplatform.org","https://dsm-dev.datadonationplatform.org",
+            "https://dsm-test.datadonationplatform.org","https://dsm-staging.datadonationplatform.org");
+
```
MocanaAtBroad: You sure about hard-coding these?

src/main/java/org/broadinstitute/dsm/DSMServer.java
MocanaAtBroad: https://github.com/broadinstitute/ddp-study-server/blob/6579631f8ccdc3cd480e621999a7fc12e94b889e/pepper-apis/src/main/java/org/broadinstitute/ddp/DataDonationPlatform.java#L602

Best bang for the buck performance boost. Allows browser to cache OPTION call result instead of preceding every API call with OPTION.

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -252,7 +269,7 @@ protected void setupCustomRouting(@NonNull Config cfg) {
     protected void updateDB(@NonNull String dbUrl) {
         logger.info("Running DB update...");
 
-        try (Connection conn = DriverManager.getConnection(dbUrl + "&verifyServerCertificate=true&useSSL=true&requireSSL=true"
+        try (Connection conn = DriverManager.getConnection(dbUrl
```
MocanaAtBroad, line 417: Did you update the JDBC URL to use the GCP options?

pom.xml
```diff
@@ -10,13 +10,45 @@
...
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+            <version>8.0.18</version>
+        </dependency>
```
MocanaAtBroad, line 51: Just checked and there is finally a version 8.0.20 out that supposedly deal with time conversions properly (as you might recall we had to set timezones in the URL and in code to make sure we always were doing UTC). I suspect this is less of an issue in DSM.

pom.xml
```diff
@@ -10,13 +10,45 @@
...
+        <dependency>
+            <groupId>com.google.cloud.sql</groupId>
+            <artifactId>mysql-socket-factory-connector-j-8</artifactId>
+            <version>1.0.15</version>
+        </dependency>
+
```
MocanaAtBroad, line 58: You probably dealt with it, but can't see the DB url so will mention it. I asked below about updating JDBC URL. Using their socket factory is one of the options in the URL. I believe same effect can be obtained when creating connection in code.

appengine/deploy/build-and-deploy.sh
zyme: Vault?   What's  that?  :)

Check  out the  [secrets here](https://console.cloud.google.com/security/secret-manager?folder=&organizationId=&project=broad-ddp-dev)

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -668,4 +685,31 @@ public static boolean isTest(@NonNull String shipperName) {
         }
         return true;
     }
+
+
+    // todo  arz fixme read from config file
```
zyme: @MocanaAtBroad to  answer  your question: yes, I'm sure that hardcoding the origin values  is a bad idea  :)

pom.xml
zyme: Yep, I used the cloudsqlproxy-style URLs.  You can check  the secrets [here](https://console.cloud.google.com/security/secret-manager?folder=&organizationId=&project=broad-ddp-dev).   in  a UI!  With secret versions!  And no weird "don't forget `.data` and `jq` stuff"...

MocanaAtBroad: I followed GAE's recommendation and setup the URL like below.
Requires granting permissions. If the proxy thing works then maybe its all good. But there might or might not be a performance penalty.
jdbc:mysql:///pepperdev?cloudSqlInstance=broad-ddp-dev:us-central1:ddp-dev&socketFactory=com.google.cloud.sql.mysql.SocketFactory&user=XXXX&password=YYYYY@&characterEncoding=UTF-8&useLegacyDatetimeCode=false&serverTimezone=UTC&useSSL=false&sessionVariables=innodb_strict_mode=on,tx_isolation='READ-COMMITTED',sql_mode='TRADITIONAL'

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -252,7 +269,7 @@ protected void setupCustomRouting(@NonNull Config cfg) {
     protected void updateDB(@NonNull String dbUrl) {
         logger.info("Running DB update...");
 
-        try (Connection conn = DriverManager.getConnection(dbUrl + "&verifyServerCertificate=true&useSSL=true&requireSSL=true"
```
zyme: Going with cloudsqlproxy, so we don't need any of the cert stuff here.

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -76,47 +89,51 @@
...
+        File vaultConfigInDeployDir = new File(GAE_DEPLOY_DIR, VAULT_DOT_CONF);
+        File vaultConfig = vaultConfigInCwd.exists()  ? vaultConfigInCwd : vaultConfigInDeployDir;
+        logger.info("Reading config values from "+vaultConfig.getAbsolutePath());
+        cfg = cfg.withFallback(ConfigFactory.parseFile(vaultConfig));
 
         if (StringUtils.isNotBlank(cfg.getString("portal.googleProjectCredentials"))) {
             System.setProperty("GOOGLE_APPLICATION_CREDENTIALS", cfg.getString("portal.googleProjectCredentials"));
         }
-
-        TransactionWrapper.configureSslProperties(cfg.getString("portal.dbSslKeyStore"),
```
zyme: Cloudsql proxy--no need for managing certs on the jdbc side.

pom.xml
zyme: Ooops, this should have been an exclusion of `com.sparkjava` from `lddp`....

pom.xml
zyme: Trying this out locally and it's causing some explosions with use of `beforeFirst()`.  Probably worth fixing and doing the upgrade...

```
2020-06-25 10:33:14,981 INFO  [JobScheduler_Worker-1] (org.broadinstitute.dsm.util.KitUtil.findSpecialBehaviorKits:530) - Checking pt N7TDWDSFIFV36KVADJJN for special behavior
2020-06-25 10:33:14,981 INFO  [JobScheduler_Worker-1] (org.broadinstitute.dsm.util.KitUtil.findSpecialBehaviorKits:530) - Checking pt PUI7OT7LTDRLBP18STRC for special behavior
2020-06-25 10:33:14,981 INFO  [JobScheduler_Worker-1] (org.broadinstitute.dsm.util.KitUtil.findSpecialBehaviorKits:530) - Checking pt 9C8LU8KT3OSBGTIPD0OW for special behavior
2020-06-25 10:33:14,982 INFO  [JobScheduler_Worker-1] (org.broadinstitute.dsm.util.KitUtil.findSpecialBehaviorKits:530) - Checking pt SDJZ2R73TGBHWGGNRCJ5 for special behavior
2020-06-25 10:33:15,020 INFO  [JobScheduler_Worker-1] (org.broadinstitute.ddp.db.TransactionWrapper.inTransaction:146) - retrieved connection...
2020-06-25 10:33:15,178 INFO  [JobScheduler_Worker-1] (org.broadinstitute.dsm.db.KitRequestShipping.getAllKitRequestsByRealm:390) - Collecting kit information
2020-06-25 10:33:15,217 INFO  [JobScheduler_Worker-1] (org.broadinstitute.ddp.db.TransactionWrapper.inTransaction:146) - retrieved connection...
2020-06-25 10:33:15,375 INFO  [JobScheduler_Worker-1] (org.broadinstitute.dsm.db.KitRequestShipping.getAllKitRequestsByRealm:470) - Got 0 participants kits in DSM DB for Osteo
2020-06-25 10:33:15,414 INFO  [JobScheduler_Worker-1] (org.broadinstitute.ddp.db.TransactionWrapper.inTransaction:146) - retrieved connection...
2020-06-25 10:33:15,572 ERROR [JobScheduler_Worker-1] (org.broadinstitute.dsm.jobs.DDPRequestJob.execute:58) - Some error occurred while doing pdf audit trail 
java.lang.RuntimeException: Error getting bookmark pdf_audit_kit
        at org.broadinstitute.dsm.util.DBUtil.getBookmark(DBUtil.java:33)
        at org.broadinstitute.dsm.util.PDFAudit.checkAndSavePDF(PDFAudit.java:18)
        at org.broadinstitute.dsm.jobs.DDPRequestJob.execute(DDPRequestJob.java:55)
        at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
        at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
Caused by: java.lang.RuntimeException: Error getting bookmark pdf_audit_kit
        at org.broadinstitute.dsm.util.DBUtil.getBookmark(DBUtil.java:53)
        at org.broadinstitute.dsm.util.DBUtil.lambda$getBookmark$0(DBUtil.java:24)
        at org.broadinstitute.ddp.db.TransactionWrapper.inTransaction(TransactionWrapper.java:150)
        at org.broadinstitute.ddp.db.TransactionWrapper.inTransaction(TransactionWrapper.java:139)
        at org.broadinstitute.dsm.util.DBUtil.getBookmark(DBUtil.java:21)
        ... 4 more
Caused by: java.sql.SQLException: Operation not allowed for a result set of type ResultSet.TYPE_FORWARD_ONLY.
        at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:129)
        at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
        at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
        at com.mysql.cj.jdbc.result.ResultSetImpl.last(ResultSetImpl.java:1711)
        at org.apache.commons.dbcp2.DelegatingResultSet.last(DelegatingResultSet.java:407)
        at org.apache.commons.dbcp2.DelegatingResultSet.last(DelegatingResultSet.java:407)
        at org.broadinstitute.dsm.util.DBUtil.getBookmark(DBUtil.java:43)
        ... 8 more
```

src/main/java/org/broadinstitute/dsm/route/AssignParticipantRoute.java
```diff
@@ -211,7 +211,8 @@ public void doNotification(String email, String message, String recordId, String
     public String getDDPParticipantId(@NonNull String participantId) {
         SimpleResult results = inTransaction((conn) -> {
             SimpleResult dbVals = new SimpleResult();
-            try (PreparedStatement stmt = conn.prepareStatement(queryForDDPParticipantId)) {
+            try (PreparedStatement stmt = conn.prepareStatement(queryForDDPParticipantId,
```
zyme, line 214: @MocanaAtBroad the pattern seems to be that with `8.0.20`, we have to set cursor modes explicitly for places that are going backwards through cursors.  I did a find/replace for areas of `rs.beforeFirst()` to try to find these.

MocanaAtBroad: Had no idea that one could do these tricks with result sets! (though this looks like someone really really does not want to write more SQL queries than absolutely necessary!)

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -668,4 +703,69 @@ public static boolean isTest(@NonNull String shipperName) {
...
+
+        public ReadinessRoute(long bootTimeoutSeconds) {
+            this.bootTimeoutSeconds = bootTimeoutSeconds;
+        }
+
+        @Override
+        public Object handle(Request req, Response res) throws Exception {
+            AtomicInteger status = new AtomicInteger(HttpStatus.SC_SERVICE_UNAVAILABLE);
+            long bootTime = Instant.now().toEpochMilli();
+            Thread waitForBoot = new Thread(() -> {
```
zyme, line 932: @MocanaAtBroad @yufengwng this turned out to be a bit more complicated than I originally thought, but it brings up something that I think we need to address in study-server as well: letting an instance start fully (allowing for liquibase to complete, which can take a while) before GAE sends requests to the instance.  Based on [this doc](https://cloud.google.com/appengine/docs/standard/java/how-instances-are-managed#startup), I think the approach here (to have `ah/start` block until things are ready) should work.  Otherwise, the default `404`  that spark returns will tell GAE "everything is fine--please start sending traffic to this instance", which is probably premature, since even the shortest deployments seem to take ~10s, without liquibase doing much.

zyme: Also, I'm really glad y'all mentioned the whole `/ah/start` thing.  It was driving me kinda bonkers.

zyme: A sampling of more recent deploys shows cleaner `ah/start` calls and their duration.  Note the lack of `404`, `500`, or `503` at top.
Screen Shot 2020-06-25 at 4 01 51 PM

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -76,47 +96,62 @@
...
+        logger.info("Using port {}", port);
+        port(port);
+        registerAppEngineStartupCallback(bootTimeoutSeconds);
         setupDB(config);
         setupRouting(config);
+        String preferredSourceIPHeader = null;
+        if (config.hasPath(ApplicationConfigConstants.PREFERRED_SOURCE_IP_HEADER)) {
+            preferredSourceIPHeader = config.getString(ApplicationConfigConstants.PREFERRED_SOURCE_IP_HEADER);
+        }
+        JettyConfig.setupJetty(preferredSourceIPHeader);
```
yufengwng: Does this `setupJetty()` call need to go before `registerAppEngineStartupCallback()`? Mounting any routes or filters will have initialized the Spark server, which is what is being done in `registerAppEngineStartupCallback()` and `setupRouting()`?

pom.xml
zyme: Gaah, I totally forgot the `UTC` url params

src/main/java/org/broadinstitute/dsm/DSMServer.java
zyme: Yep!

src/main/java/org/broadinstitute/dsm/DSMServer.java
zyme: Looking at the  DSM logs (with the real IP), it doesn't look like the order matters here.  This  order of calls still produces  the correct IP address in DSM logs.

src/main/java/org/broadinstitute/dsm/route/LoggingFilter.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.log4j.MDC;
+import spark.Filter;
+import spark.Request;
+import spark.Response;
+
+/**
+ * Captures client IP and name of user from token, adding
+ * them to MDC for logging.
+ */
+public class LoggingFilter implements Filter {
```
zyme, line 24: Was surprised  that we  aren't logging IP and JWT user id info--so I added it, since this is an important part of compliance.

Screen Shot 2020-06-29 at 11 46 30 AM

src/main/java/org/broadinstitute/dsm/DSMServer.java
yufengwng: I still think it's a bit risky. From my reading of Spark's internals, the embedded server is initialized in a separate thread, so I suspect we might just be getting lucky that our custom factory is getting added before the thread goes through it's paces.

pom.xml
```diff
@@ -145,79 +211,27 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.2</version>
+                <version>3.8.0</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <release>11</release>
```
zyme, line 266: Woot, java11!

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -76,47 +97,66 @@
...
+
+
+        logger.info("Using port {}", port);
+        port(port);
+
+        registerAppEngineStartupCallback(bootTimeoutSeconds);
         setupDB(config);
-        setupRouting(config);
+        // don't run superclass routing--it won't work with JettyConfig changes for capturing proper IP address in GAE
+        setupCustomRouting(config);
```
zyme, line 197: @pegahtah this change fixed the classnotfoundexception stuff.  Can you try this locally?

pom.xml
MocanaAtBroad: I don't think you need sqlproxy to run in GAE. Just need to grant permissions to the service. Read somewhere (don't have the reference, sorry) that sqlproxy introduced latency.

pom.xml
```diff
@@ -10,15 +10,46 @@
...
+            <groupId>com.google.cloud.sql</groupId>
+            <artifactId>mysql-socket-factory-connector-j-8</artifactId>
+            <version>1.0.15</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.sparkjava</groupId>
+            <artifactId>spark-core</artifactId>
+            <version>2.8.0</version>
+        </dependency>
```
yufengwng: Looks like a repeat of the one on L15.

yufengwng (commented):
Overall looks good to me. Somehow I don't have the "Resolve conversation" button when viewing this PR in GitHub. Please feel free to close out comments.

pom.xml
zyme: Sorry for the confusion--when I said "cloudsqlproxy-style", I meant the urls like `cloudSqlInstance=broad-ddp-dev:us-central1:ddp-dev`.  So not the cloudsqlproxy per se, but urls that reference gcp project, zone, and cloudsql instance name instead of IP.

pom.xml
zyme: Double import, just to make double sure 🤦

appengine/deploy/build-and-deploy.sh
```diff
@@ -0,0 +1,26 @@
...
+mvn -f ../../pom.xml dependency:copy-dependencies -DoutputDirectory=./appengine/deploy/lib
+cp ../../target/DSMServer.jar .
+cp ../../src/main/resources/log4j.xml .
+
+echo "Downloading and configuring tcell"
+gsutil cat gs://ddp-tcell/tcell-1.11.0.tar.gz | tar -xf -
+gcloud --project=${PROJECT_ID} secrets versions access latest --secret="study-manager-tcell" >  tcell/tcell_agent.config
+
+# deploy to gae
+gcloud app deploy -q --stop-previous-version --promote --project ${PROJECT_ID} StudyManager.yaml
```
MocanaAtBroad: One thing you might want to consider here is deleting older versions of deployed service. There is an overall limit in in the GCP project. On the Pepper side, right after we deploy, we try to remove any older versions. 
What service count look like right now in ddp-dev:
Screen Shot 2020-07-23 at 7 11 41 AM

How we deal with it in CircleCi:
https://github.com/broadinstitute/ddp-study-server/blob/7c7843b080b0fd9289178c38ea9477025cc99a18/.circleci/config.yml#L294

Removing versions sometime simply fails. If that happens we just let it go and try to remove it again next time there is a deployment.

SimoneMaiwald:
going to close old PRs now

PR #35, opened 2020-07-15
Ddp 4769 batch kit endpoint

pegahtah:
I create a new end point per ticket DDP-4769 so Pepper can query DSM for the kit status of a list of participants.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/route/BatchKitsRoute.java
+++ b/src/main/java/org/broadinstitute/dsm/route/BatchKitsRoute.java
@@ -1,0 +1,55 @@
+package org.broadinstitute.dsm.route;
+
+import com.google.gson.Gson;
+import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.dsm.db.DDPInstance;
+import org.broadinstitute.dsm.db.KitRequestShipping;
+import org.broadinstitute.dsm.statics.RequestParameter;
+import org.broadinstitute.dsm.util.SystemUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import spark.Request;
+import spark.Response;
+import spark.Route;
+
+import javax.servlet.http.HttpServletRequest;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class BatchKitsRoute implements Route {
+    private static final Logger logger = LoggerFactory.getLogger(BatchKitsRoute.class);
+
+    @Override
+    public Object handle(Request request, Response response) throws Exception {
+        String study = request.params(RequestParameter.REALM);
+        String ids = null;
+        if (StringUtils.isNotBlank(study)) {
+            HttpServletRequest rawRequest = request.raw();
+            String content = SystemUtil.getBody(rawRequest);
+            ids = (String) (new Gson().fromJson(content, new HashMap<String, String[]>().getClass())).get("participantIds");
+            String[] ddpParticipantIds = new Gson().fromJson(ids, String[].class);
+            DDPInstance ddpInstance = DDPInstance.getDDPInstance(study);
+            if (ddpInstance != null) {
+                if (ddpParticipantIds != null && ddpParticipantIds.length != 0) {
+                    Map<String, List<KitRequestShipping>> kitRequests = KitRequestShipping.getKitRequests(ddpInstance.getName());
+                    Map<String, List<KitRequestShipping>> results = new HashMap<>();
+                    for (String ddpParticipantId : ddpParticipantIds) {
+                        results.put(ddpParticipantId, kitRequests.get(ddpParticipantId));
+                    }
+                    logger.info("Sending a list of "+results.size()+" KitRequestShippings for study "+study);
+                    return results;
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/dsm/route/BatchKitsRoute.java
```diff
@@ -0,0 +1,55 @@
...
+            String content = SystemUtil.getBody(rawRequest);
+            ids = (String) (new Gson().fromJson(content, new HashMap<String, String[]>().getClass())).get("participantIds");
+            String[] ddpParticipantIds = new Gson().fromJson(ids, String[].class);
+            DDPInstance ddpInstance = DDPInstance.getDDPInstance(study);
+            if (ddpInstance != null) {
+                if (ddpParticipantIds != null && ddpParticipantIds.length != 0) {
+                    Map<String, List<KitRequestShipping>> kitRequests = KitRequestShipping.getKitRequests(ddpInstance.getName());
+                    Map<String, List<KitRequestShipping>> results = new HashMap<>();
+                    for (String ddpParticipantId : ddpParticipantIds) {
+                        results.put(ddpParticipantId, kitRequests.get(ddpParticipantId));
```
zyme: Is client  expecting to see a subset of the values that it posted, or should this iterate over `ids` so that the response includes all the ids that we passed in, with `null` or `[]` as the value for kits if the participant doesn't have any kits?

zyme: Any performance concerns or memory usage concerns here if we end up with 40,000 kits in this list?

pegahtah: I assumed the second scenario here. Any thoughts @yufengwng ?

pegahtah: I think this works per # of participants not kits, which is good

zyme: I'm not too concerned about it.  It'll take a while to get up to that volume, and it  doesn't look like the objects themselves use a ton of space.  Probably not worth the time to optimize it now, especially since the GBF integration is more of a blocker.

src/main/java/org/broadinstitute/dsm/route/BatchKitsRoute.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public Object handle(Request request, Response response) throws Exception {
+        String study = request.params(RequestParameter.REALM);
+        String ids = null;
+        if (StringUtils.isNotBlank(study)) {
+            HttpServletRequest rawRequest = request.raw();
+            String content = SystemUtil.getBody(rawRequest);
+            ids = (String) (new Gson().fromJson(content, new HashMap<String, String[]>().getClass())).get("participantIds");
+            String[] ddpParticipantIds = new Gson().fromJson(ids, String[].class);
```
yufengwng: We should probably put a limit on how big a batch can be. Maybe max 100 ids?

src/main/java/org/broadinstitute/dsm/route/BatchKitsRoute.java
```diff
@@ -0,0 +1,55 @@
...
+            DDPInstance ddpInstance = DDPInstance.getDDPInstance(study);
+            if (ddpInstance != null) {
+                if (ddpParticipantIds != null && ddpParticipantIds.length != 0) {
+                    Map<String, List<KitRequestShipping>> kitRequests = KitRequestShipping.getKitRequests(ddpInstance.getName());
+                    Map<String, List<KitRequestShipping>> results = new HashMap<>();
```
yufengwng: Can we use a different JSON structure here? Having the participant id as a key is unwieldy to work with in JSON. Can we get a list of objects instead?

src/main/java/org/broadinstitute/dsm/route/BatchKitsRoute.java
```diff
@@ -0,0 +1,55 @@
...
+                    for (String ddpParticipantId : ddpParticipantIds) {
+                        results.put(ddpParticipantId, kitRequests.get(ddpParticipantId));
+                    }
+                    logger.info("Sending a list of "+results.size()+" KitRequestShippings for study "+study);
+                    return results;
+                }
+                else {
+                    return null;
```
yufengwng: Can we return an empty list instead?

src/main/java/org/broadinstitute/dsm/route/BatchKitsRoute.java
yufengwng: I think we want something similar to first scenario. If participant id is not found, there won't be an object for it in the final result list, rather than having a placeholder in it's place.

yufengwng: I think if we cap how big a batch can be that should help in managing the volume.

zyme (approved):
Looks good to me1

yufengwng (approved):
I think this is ready to go.

PR #38, opened 2020-07-27
DDP-4815 Placing orders with CareEvolve for COVID-19

zyme:
DDP-4815 is about setting up a utility method for placing orders in CareEvolve for COVID-19. Covid19OrderRegistrar is where most of the action is for this PR--just read the auth settings from SM, create a new Order, fill in details from ES, wrap it in a Message, and off it goes to CareEvolve.

Once this service is blessed, next steps would be to call the service when shipping confirmation and labels come back from GBF.  At that time, we'd need to query elastic to read a bit more data from study-server such as gender and ethnicity.

Start at the[ test for this](https://github.com/broadinstitute/ddp-study-manager/pull/38/files#diff-74799b3e87c38651e190793dbe7d8ecc), which is currently disabled but should pass.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
+++ b/src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
@@ -76,9 +76,6 @@ public class Covid19OrderRegistrar {
         queryConditions.put("ES", " AND profile.hruid = '" + participantHruid +"'");
         List<ParticipantWrapper> participants = ParticipantWrapper.getFilteredList(instance, queryConditions);
 
-        // "activityCode" -> "BASELINE_COVID"
-        // data.get("activities") DOB SEX RACE
-
         if (participants.size() == 1) {
             ParticipantWrapper participant = participants.iterator().next();
             JsonObject data = participant.getDataAsJson();
--- a/src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
+++ b/src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
@@ -129,7 +129,7 @@ public class Covid19OrderRegistrar {
 
                     // mappings  for  race, ethnicity, and  sex are custom  for the  Ellkay API
                     // and are controlled  by the Ellkay integration
-                    String race = "2131-1";
+                    String race = "2131-1"; // code for "other"
                     String ethnicity = "U";
                     String sex = "U";
                     for (int i = 0; i < baselineCovidAnswers.size(); i++) {
@@ -169,11 +169,10 @@ public class Covid19OrderRegistrar {
                         }  else if ("RACE".equals(questionStableId)) {
                             JsonArray races = answer.getAsJsonArray("answer");
 
-                            if (races.size() > 0) {
-                                // CareEvolve only supports a  single race per order, so we'll pick
-                                // one of  the selected values at random
+                            if (races.size() == 1) {
+                                // CareEvolve only supports a  single race per order
                                 int randomIndex = ThreadLocalRandom.current().nextInt(races.size());
-                                race = races.get(randomIndex).getAsString();
+                                race = races.get(0).getAsString();
 
                                 if ("ASIAN".equals(race)) {
                                     race = "2028-9";
@@ -190,6 +189,10 @@ public class Covid19OrderRegistrar {
                                 } else {
                                     logger.error("Could not map race " + race + " to Ellkay code; will default to other for " + baselineCovidActivity.get("guid"));
                                 }
+                            } else {
+                                // if there's no value or if multiple values are selected,
+                                // use the code for "other"
+                                race = "2131-1";
                             }
                         }
                         else if ("ETHNICITY".equals(questionStableId)) {
--- a/src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
+++ b/src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
@@ -7,7 +7,6 @@ import java.time.Instant;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ThreadLocalRandom;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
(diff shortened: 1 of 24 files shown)


src/main/java/org/broadinstitute/dsm/careevolve/Provider.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @SerializedName("FirstName")
+    private String firstName;
+
+    @SerializedName("LastName")
+    private String lastName;
+    private final String npi;
+
+    @SerializedName("NPI")
+    private String NPI;
```
yufengwng: This needs a bit of cleanup. Right now constructor is assigning to `npi`.

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -0,0 +1,145 @@
...
+ */
+public class Covid19OrderRegistrar {
+
+    private static final Logger logger = LoggerFactory.getLogger(Covid19OrderRegistrar.class);
+
+    private final String endpoint;
+
+    private final String careEvolveAccount;
+
+    private final Provider provider;
```
yufengwng, line 41: Looks like the provider right now is the TestBoston's PI. It feels like other studies might want their own provider, but this should be fine for now since only TestBoston is using this?

src/main/java/org/broadinstitute/dsm/careevolve/Order.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.List;
+
+import com.google.gson.annotations.SerializedName;
+
+public class Order {
+
+    @SerializedName("AOEs")
+    private List<AOE> aoes = new ArrayList<>();
+
+    @SerializedName("collection")
```
yufengwng: Do these need to be capitalized as well? (`Collection`, `Patient`, etc). Also, do we need to hardcode things like `TestCode`?

yufengwng (commented):
Makes sense so far. I got a bit confused about the structure of the request/response payloads since those doesn't quite match what Ellkay has in their PDF, but I'll assume this works.

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
zyme: I'm hoping so!  If that changes, we can always new one up with a different provider, which I'm thinking might also be a different CE account.

src/main/java/org/broadinstitute/dsm/careevolve/Order.java
zyme: Yep!  Thanks for pointing that out.

zyme:
Update: addressed the feedback here, as well as adding collection time.

zyme:
@pegahtah @yufengwng now that we've gotten a round trip through Ellkay and back, can  you please re-review?

Please  see  the example HL7 [in slack](https://broadinstitute.slack.com/archives/G012Y2S45KN/p1597191983003100).

zyme:
@pegahtah @yufengwng I've updated this to include parsing of collection date and mappings for race, ethnicity, and sex.  Can y'all please re-review?

Also, @pegahtah what is the best practice for DSM branch management in terms of what is on prod?  We should update various branches and docs to make sure we know what is currently on prod, and how this PR should flow to get onto prod.

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -0,0 +1,303 @@
...
+                if (address != null) {
+                    Address careEvolveAddress = toCareEvolveAddress(address);
+
+                    JsonObject profile = data.get("profile").getAsJsonObject();
+                    String patientId = profile.get("guid").getAsString();
+
+                    String firstName = profile.get("firstName").getAsString();
+                    String lastName = profile.get("lastName").getAsString();
+
+                    List<AOE> aoes = AOE.forTestBoston(null, kitId);
```
yufengwng: Should we pass in the user guid here? Or should we remove that AOE?

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -0,0 +1,303 @@
...
+
+                        if ("SAMPLE_COLLECT_DATE".equals(questionStableId)) {
+                            JsonObject dateFields = latestKitAnswer.get("dateFields").getAsJsonObject();
+                            collectionDateTimeStr.append(dateFields.get("month")).append("/")
+                                    .append(dateFields.get("day")).append("/")
+                                    .append(dateFields.get("year"));
+                        } else if ("SAMPLE_COLLECT_TIME".equals(questionStableId)) {
+                            JsonArray timeFields = latestKitAnswer.getAsJsonArray("answer").get(0).getAsJsonArray();
+                            for (int timeFieldIdx = 0; timeFieldIdx < timeFields.size(); timeFieldIdx++) {
+                                collectionDateTimeStr.append(" ").append(timeFields.get(timeFieldIdx).getAsString());
```
yufengwng: To be on the safe side, how about we concat the date and time parts _after_ the loop? I'm not sure the ordering is guaranteed and that time might come before date?

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -0,0 +1,303 @@
...
+                    // and are controlled  by the Ellkay integration
+                    String race = "2131-1"; // code for "other"
+                    String ethnicity = "U";
+                    String sex = "U";
+                    for (int i = 0; i < baselineCovidAnswers.size(); i++) {
+                        JsonObject answer = baselineCovidAnswers.get(i).getAsJsonObject();
+                        String questionStableId = answer.get("stableId").getAsString();
+
+                        if ("DOB".equals(questionStableId)) {
+                            JsonObject dateFields = answer.get("dateFields").getAsJsonObject();
```
yufengwng, line 151: For date answers where we have all the pieces, we do export a full ISO date. Maybe we can use that?
```
                {
                  "date" : "2002-01-01",
                  "dateFields" : {
                    "month" : 1,
                    "year" : 2002,
                    "day" : 1
                  },
                  "stableId" : "DOB",
                  "questionType" : "DATE"
                },
```

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -0,0 +1,303 @@
...
+                            }  else {
+                                logger.error("Could not map sex " + sex + " to Ellkay code; will default to unknown for " + baselineCovidActivity.get("guid"));
+                            }
+
+                        }  else if ("RACE".equals(questionStableId)) {
+                            JsonArray races = answer.getAsJsonArray("answer");
+
+                            if (races.size() == 1) {
+                                // CareEvolve only supports a  single race per order
+                                int randomIndex = ThreadLocalRandom.current().nextInt(races.size());
```
yufengwng: Looks like we can remove this line.

yufengwng (commented):
Looks good so far!

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
yufengwng: It looks like our data exporter is always iterating by question definition order, and date question is defined before time question, so I guess this should be okay.

zyme: Will move the  date + time  out   of  the  loop...

zyme:
One last re-review?

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -0,0 +1,302 @@
...
+                    // mappings  for  race, ethnicity, and  sex are custom  for the  Ellkay API
+                    // and are controlled  by the Ellkay integration
+                    String race = "2131-1"; // code for "other"
+                    String ethnicity = "U";
+                    String sex = "U";
+                    for (int i = 0; i < baselineCovidAnswers.size(); i++) {
+                        JsonObject answer = baselineCovidAnswers.get(i).getAsJsonObject();
+                        String questionStableId = answer.get("stableId").getAsString();
+
+                        if ("DOB".equals(questionStableId)) {
```
yufengwng, line 150: D'oh! I just realized we're not exporting the birth_date in profile to ES. I'm making a change in TestBoston to copy `DOB` answer to birth_date so we have that there. Might be easier to grab from that instead of from the activity answers itself.

yufengwng (commented):
Looks good to me!

yufengwng (approved):

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
yufengwng: Just realized that we _are_ exporting birth date already under `dsm.dateOfBirth`. I'll recommend grabbing from that but the current way is fine too.

PR #42, opened 2020-08-31
testResults-pubsub-integrartion

pegahtah:


src/main/java/org/broadinstitute/dsm/model/birch/TestBostonResult.java
```diff
@@ -0,0 +1,7 @@
+package org.broadinstitute.dsm.model.birch;
+
+public class TestBostonResult {
+    public String date;
+    public String sampleId;
+    public String result;
```
pegahtah: This will be changed in the next push, I was testing with this json and I pushed the wrong code. But you get the idea

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -89,6 +88,10 @@
     public static final String ADDITIONAL_CRON_EXPRESSION = "AdditionalCronExpression";
     public static final String GCP_PATH_TO_SERVICE_ACCOUNT = "portal.googleProjectCredentials";
 
+    public static final String PUBSUB_BIRCH_TOPIC = "";
+    public static final String PUBSUB_PROJECT_ID = "hl7-dev-274911";
+    public static final String PUBSUB_BIRCH_SUBSCRIPTION = "pegah-results-sub";
```
pegahtah: Using one of my own old test subscriptions to a fake topic I created at the beginning of birch for testing.

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUpJob.java
```diff
@@ -0,0 +1,92 @@
...
+            // Instantiate an asynchronous message receiver.
+            MessageReceiver receiver =
+                    (PubsubMessage message, AckReplyConsumer consumer) -> {
+                        // Handle incoming message, then ack the received message.
+                        System.out.println("Id: " + message.getMessageId());
+                        System.out.println("Data: " + message.getData().toStringUtf8());
+                        writeResultsIntoDB(message);
+                        consumer.ack();
+                    };
+
```
yufengwng: How often does this run, what's the cron schedule right now? Do we want to setup a long-running listener rather than spinning up/down a subscriber through a job?

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUpJob.java
```diff
@@ -0,0 +1,92 @@
...
+            Subscriber subscriber = null;
+            try {
+                subscriber = Subscriber.newBuilder(subscriptionName, receiver).build();
+                // Start the subscriber.
+                subscriber.startAsync().awaitRunning();
+                logger.info("Listening for messages on %s:\n", subscriptionName.toString());
+                // Allow the subscriber to run for 30s unless an unrecoverable error occurs.
+                subscriber.awaitTerminated(1L, TimeUnit.MINUTES);
```
yufengwng: Interesting. Since we haven't called stop before this line, this trick is so to timeout the subscriber after a minute?

pegahtah: based on GCP's guide I think this should do the trick.

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUpJob.java
pegahtah: Good question, I run it every 30 minutes now, but we need to figure it out based on how "live" they want the results. In the current configuration,  does this mean we have to set the messages on the pubsub's topic to also have a 30 minute ack deadline time?

zyme: I'd prefer  a long-term listener instead of startting/stopping subscribers inside  of scheduled  jobs.  It's more consistent with  our  other pubsub  methods.

We  can flow control this  with a single thread as well.

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUpJob.java
```diff
@@ -0,0 +1,92 @@
...
+                    stmt.setString(3, finalTestBostonResult.sampleId);
+                    stmt.executeUpdate();
+                }
+            }
+            catch (SQLException ex) {
+                dbVals.resultException = ex;
+            }
+            return dbVals;
+        });
+
```
zyme: Can we log cases where we are *changing* the result (instead of going from zero results  to one result)?

pegahtah: A little confused here, do you mean not change them, just log instead?

zyme: We should update the value and log that it has been updated.

pegahtah: 👍

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUpJob.java
```diff
@@ -0,0 +1,139 @@
...
+                "        where request.dsm_kit_request_id = kit.dsm_kit_request_id " +
+                "        and request.ddp_instance_id = realm.ddp_instance_id " +
+                "        and (eve.ddp_instance_id = request.ddp_instance_id " +
+                "        and eve.kit_type_id = request.kit_type_id) " +
+                "        and eve.event_type = \"RESULT\" " + // that's the change from the original query
+                "        and request.external_order_number = ?"; // that's the change from the original query
+
+        KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(query, testBostonResult.getSampleId());
+        if (kitDDPNotification != null) {
+            EventUtil.triggerDDPWithTestResult(kitDDPNotification, testBostonResult);
```
pegahtah: @SimoneMaiwald

src/main/java/org/broadinstitute/dsm/model/TestBostonEvent.java
```diff
@@ -0,0 +1,14 @@
+package org.broadinstitute.dsm.model;
+
+import lombok.NonNull;
+import org.broadinstitute.ddp.handlers.util.Event;
+import org.broadinstitute.dsm.model.birch.TestBostonResult;
+
+public class TestBostonEvent extends Event {
```
pegahtah: @SimoneMaiwald

yufengwng: Maybe we should name this a bit more general, say, `TestResultEvent`?

src/main/java/org/broadinstitute/dsm/model/TestBostonEvent.java
```diff
@@ -0,0 +1,14 @@
...
+    TestBostonResult result;
```
yufengwng: I thought this was going to be called `eventData`, where idea is that other events might also have data as well. Should Pepper expect a property called `eventData` or `result`?

src/main/java/org/broadinstitute/dsm/model/birch/TestBostonResult.java
```diff
@@ -0,0 +1,14 @@
+package org.broadinstitute.dsm.model.birch;
+
+import lombok.Data;
+
+@Data
+public class TestBostonResult {
+    private String orderMessageId;
+    private String institutionId;
+    private String sampleId;
```
yufengwng, line 13: The name `sampleId` seems a bit foreign to how DSM/Pepper usually communicate. This is the `kitRequestId` right?

src/main/java/org/broadinstitute/dsm/model/birch/TestBostonResult.java
```diff
@@ -0,0 +1,14 @@
...
+    private String result;
+    private String reason;
+    private String timeCompleted;
+
```
yufengwng, line 21: How are we thinking of handling potential duplicates? We can do de-dup in Pepper, but we'll need to get the `is_corrected` in here as well.

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -294,10 +306,51 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
+            ProjectSubscriptionName subscriptionName =
+                    ProjectSubscriptionName.of(projectId, subscriptionId);
+
+            // Instantiate an asynchronous message receiver.
+            MessageReceiver receiver =
+                    (PubsubMessage message, AckReplyConsumer consumer) -> {
+                        // Handle incoming message, then ack the received message.
+                        System.out.println("Id: " + message.getMessageId());
+                        System.out.println("Data: " + message.getData().toStringUtf8());
+                        PubSubLookUp.processCovidTestResults(message);
```
yufengwng: For good measure, perhaps wrap this in a try/catch and call `nack()` on an exception?

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,144 @@
...
+    private static DSMTestResult[] getLatestKitTestResults(TestBostonResult testBostonResult) {
+        String externalOrderNumber = testBostonResult.getSampleId();
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult dbVals = new SimpleResult();
+            try (PreparedStatement stmt = conn.prepareStatement(SELECT_LATEST_RESULT_QUERY)) {
+                stmt.setString(1, externalOrderNumber);
+                try (ResultSet rs = stmt.executeQuery()) {
+                    if (rs.next()) {
+                        String resultArray = rs.getString("test_result");
+                        DSMTestResult[] dsmTestResultsArray = new Gson().fromJson(resultArray, DSMTestResult[].class);
```
yufengwng, line 92: Hmmm interesting, so idea is to store it as a json array instead of one per database row?

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,144 @@
...
+        TestBostonResult testBostonResult = new Gson().fromJson(data, TestBostonResult.class);
+        if (shouldWriteResultIntoDB(testBostonResult)) {
+            writeResultsIntoDB(testBostonResult);
+            tellPepperAboutTheNewResults(testBostonResult);// notify pepper if we update DB
+        }
+    }
+
+    private static boolean shouldWriteResultIntoDB(TestBostonResult testBostonResult) {
+        DSMTestResult[] dsmTestResultArray = getLatestKitTestResults(testBostonResult);
+        if (dsmTestResultArray == null) {
```
yufengwng: How about making it `not null and length > 0` for good measure? Should also help simplify the `if` condition below.

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,144 @@
...
+            return true;
+        }
+        DSMTestResult dsmTestResult = null;
+        if (dsmTestResultArray != null && dsmTestResultArray.length > 0) {
+            dsmTestResult = dsmTestResultArray[dsmTestResultArray.length - 1];
+        }
+
+        //corrected result -> assuming corrected results are always changed
+        if (dsmTestResult != null && !testBostonResult.isCorrected() && dsmTestResult.isCorrected) {
```
yufengwng: Wait, so this means if existing result is corrected and new result is NOT corrected, we let it pass and save it? 🤔

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,144 @@
...
+        }
+        //duplicate result
+        if (testBostonResult.isCorrected() == dsmTestResult.isCorrected
+                && StringUtils.isNotBlank(dsmTestResult.result) && dsmTestResult.result.equals(testBostonResult.getResult())
+                && StringUtils.isNotBlank(dsmTestResult.date) && dsmTestResult.date.equals(testBostonResult.getTimeCompleted())) {
+            return false;
+        }
+        // weird result
+        if (!testBostonResult.isCorrected()
+                && StringUtils.isNotBlank(dsmTestResult.result) && !dsmTestResult.result.equals(testBostonResult.getResult()) ) {
```
yufengwng: Should this just be `new result not corrected`? So that if new result is not corrected and we have existing result, we error?

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,144 @@
...
+
+        if (results.resultException != null) {
+            throw new RuntimeException("", results.resultException);
+        }
+        DSMTestResult[] dsmTestResult = (DSMTestResult[]) results.resultValue;
+        return dsmTestResult;
+    }
+
+    public static void writeResultsIntoDB(TestBostonResult testBostonResult) {
+        DSMTestResult[] dsmTestResultArray = getLatestKitTestResults(testBostonResult);
```
yufengwng, line 114: One optimization is to just do one single query for latest kit results. We can do the query in the parent process method and pass that array into the `shouldWriteResultIntoDB()` and this method.

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -294,10 +306,51 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
         setupJobs(cfg, kitUtil, notificationUtil, eventUtil, container, receiver);
+        
+        setupPubSub();
 
         logger.info("Finished setting up DSM custom routes and jobs...");
     }
 
+    private void setupPubSub() {
+        String projectId = "hl7-dev-274911";
+        String subscriptionId = "pegah-results-sub";
```
SimoneMaiwald: you could use the static strings =)
maybe we could also move them to the config file

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -294,10 +306,51 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
+        try {
+            ProjectSubscriptionName subscriptionName =
+                    ProjectSubscriptionName.of(projectId, subscriptionId);
+
+            // Instantiate an asynchronous message receiver.
+            MessageReceiver receiver =
+                    (PubsubMessage message, AckReplyConsumer consumer) -> {
+                        // Handle incoming message, then ack the received message.
+                        System.out.println("Id: " + message.getMessageId());
+                        System.out.println("Data: " + message.getData().toStringUtf8());
```
SimoneMaiwald: could we change the system.out to logger.info please

src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -154,6 +160,7 @@ public KitRequestShipping(String dsmKitRequestId, String dsmKitId, String easypo
     }
 
     // shippingId = ddp_label !!!
+    //todo pegah add testResult and resultDate
```
SimoneMaiwald, line 168: ?

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,144 @@
...
+import com.google.gson.Gson;
+import com.google.pubsub.v1.PubsubMessage;
+import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.ddp.db.SimpleResult;
+import org.broadinstitute.dsm.model.KitDDPNotification;
+import org.broadinstitute.dsm.model.birch.DSMTestResult;
+import org.broadinstitute.dsm.model.birch.TestBostonResult;
+import org.broadinstitute.dsm.util.EventUtil;
+import org.json.JSONString;
+import org.junit.Test;
```
SimoneMaiwald: could we remove unused imports please

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
SimoneMaiwald: will throw NPE

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
SimoneMaiwald: please use for dsmTestResult.isCorrected getter method dsmTestResult.isCorrected()

SimoneMaiwald: same for dsmTestResult.result and dsmTestResult.date

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,144 @@
...
+        if (results.resultException != null) {
+            throw new RuntimeException("", results.resultException);
+        }
+        DSMTestResult[] dsmTestResult = (DSMTestResult[]) results.resultValue;
+        return dsmTestResult;
+    }
+
+    public static void writeResultsIntoDB(TestBostonResult testBostonResult) {
+        DSMTestResult[] dsmTestResultArray = getLatestKitTestResults(testBostonResult);
+        String query = "UPDATE ddp_kit SET  test_result = ? WHERE dsm_kit_id <> 0 and  dsm_kit_id  in ( select  dsm_kit_id from (select * from ddp_kit) as something where kit_label= ?  )";
```
SimoneMaiwald: move query to static

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,144 @@
...
+                && StringUtils.isNotBlank(dsmTestResult.result) && !dsmTestResult.result.equals(testBostonResult.getResult()) ) {
+
+            logger.error("A new result for sample id "+testBostonResult.getSampleId() +" that doesn't match the previous one. Date of the new result: "+testBostonResult.getTimeCompleted());
+            return false;
+        }
+        return true;
+    }
+
+    private static void tellPepperAboutTheNewResults(TestBostonResult testBostonResult) {
+        String query = "select   eve.event_name,   eve.event_type,   request.ddp_participant_id,   request.dsm_kit_request_id,   realm.ddp_instance_id,   realm.instance_name,   realm.base_url,   realm.auth0_token,   realm.notification_recipients,   realm.migrated_ddp,   kit.receive_date,   kit.scan_date   from   ddp_kit_request request,   ddp_kit kit,   event_type eve,   ddp_instance realm   where request.dsm_kit_request_id = kit.dsm_kit_request_id   and request.ddp_instance_id = realm.ddp_instance_id   " +
```
SimoneMaiwald: move query into static and remove weird spacing

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
SimoneMaiwald: do we really want that as an error. maybe warning would be enough

src/main/java/org/broadinstitute/dsm/model/birch/DSMTestResult.java
```diff
@@ -0,0 +1,13 @@
+package org.broadinstitute.dsm.model.birch;
+
+public class DSMTestResult {
+    public boolean isCorrected;
+    public String result;
+    public String date;
```
SimoneMaiwald: add getter for them  (and setter if you need…)

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,144 @@
...
+
+    private static DSMTestResult[] getLatestKitTestResults(TestBostonResult testBostonResult) {
+        String externalOrderNumber = testBostonResult.getSampleId();
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult dbVals = new SimpleResult();
+            try (PreparedStatement stmt = conn.prepareStatement(SELECT_LATEST_RESULT_QUERY)) {
+                stmt.setString(1, externalOrderNumber);
+                try (ResultSet rs = stmt.executeQuery()) {
+                    if (rs.next()) {
+                        String resultArray = rs.getString("test_result");
```
SimoneMaiwald: use string from src/main/java/org/broadinstitute/dsm/statics/DBConstants.java

src/main/resources/liquibase/036-updateDB.xml
```diff
@@ -0,0 +1,17 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet author="pegah" id="covid-testResults">
+        <addColumn tableName="ddp_kit">
+            <column name="test_result" type="VARCHAR(100)" afterColumn="needs_approval">
+            </column>
+        </addColumn>
+        <addColumn tableName="ddp_kit">
+            <column name="result_date" type="VARCHAR(100)" afterColumn="test_result">
```
SimoneMaiwald: I only see code to write into test_result. is result_date needed?

src/main/java/org/broadinstitute/dsm/DSMServer.java
pegahtah: I'll remove them after we test today anyways

src/main/java/org/broadinstitute/dsm/DSMServer.java
pegahtah: Oh thanks, I was wondering what is the best way to nack Pubsub

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
pegahtah: Great catch! thanks!!

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
pegahtah: @yufengwng I was thinking if we get the same result as corrected and as not, do we care?

@SimoneMaiwald Wouldn't we wanna know if there are dual results for a participant?

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
pegahtah: Yes because there is the possibility of having more than one per each kit, right? (considering revised and corrected ones )

src/main/resources/liquibase/036-updateDB.xml
pegahtah: nope gonna get removed ( I think I forgot to push that )

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -295,9 +304,56 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
+            MessageReceiver receiver =
+                    (PubsubMessage message, AckReplyConsumer consumer) -> {
+                        // Handle incoming message, then ack the received message.
+                        try {
+                            PubSubLookUp.processCovidTestResults(message);
+                            logger.info("Processing the message finished");
+                            consumer.ack();
+
+                        }catch(Exception ex){
+                            logger.info("about to nack the message");
```
yufengwng: It's better to log the exception instead of printing stacktrace directly.
```suggestion
                            logger.info("about to nack the message", ex);
```

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,156 @@
...
+        DSMTestResult[] dsmTestResultArray = getLatestKitTestResults(testBostonResult);
+        if (dsmTestResultArray == null || dsmTestResultArray.length == 0) {
+            return true;
+        }
+        DSMTestResult dsmTestResult = null;
+        dsmTestResult = dsmTestResultArray[dsmTestResultArray.length - 1];
+
+        //corrected result -> assuming corrected results are always changed
+        if (dsmTestResult != null && !testBostonResult.isCorrected() && dsmTestResult.isCorrected()) {
+            return false;//should we error?
```
yufengwng, line 50: Should probably add a log statement here. I'll vote for `logger.error()`.

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
```diff
@@ -0,0 +1,156 @@
...
+            array = Arrays.copyOf(dsmTestResultArray, dsmTestResultArray.length + 1); //create new array from old array and allocate one more element
+            array[array.length - 1] = newDsmTestResult;
+        }
+        String finalArray = new Gson().toJson(array);
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult dbVals = new SimpleResult();
+            try (PreparedStatement stmt = conn.prepareStatement(UPDATE_TEST_RESULT)) {
+                if (StringUtils.isNotBlank(testBostonResult.getResult())
+                        && StringUtils.isNotBlank(testBostonResult.getTimeCompleted())
+                        && StringUtils.isNotBlank(testBostonResult.getSampleId())) {
```
yufengwng, line 130: Might be better to move this check up. No point in opening transaction if we don't pass this check.

src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -44,7 +44,7 @@
             "WHERE req.ddp_instance_id = ddp_site.ddp_instance_id AND req.kit_type_id = kt.kit_type_id) AS request " +
             "LEFT JOIN (SELECT * FROM (SELECT kit.dsm_kit_request_id, kit.dsm_kit_id, kit.kit_complete, kit.label_url_to, kit.label_url_return, kit.tracking_to_id, " +
             "kit.tracking_return_id, kit.easypost_tracking_to_url, kit.easypost_tracking_return_url, kit.easypost_to_id, kit.easypost_shipment_status, kit.scan_date, kit.label_date, kit.error, kit.message, " +
-            "kit.receive_date, kit.deactivated_date, kit.easypost_address_id_to, kit.deactivation_reason, tracking.tracking_id, kit.kit_label, kit.express FROM ddp_kit kit " +
+            "kit.receive_date, kit.deactivated_date, kit.easypost_address_id_to, kit.deactivation_reason, tracking.tracking_id, kit.kit_label, kit.express, kit.test_result, kit.result_date FROM ddp_kit kit " +
```
yufengwng: Doesn't look like we have a `kit.result_date` column? We also need to remove the `private String resultDate` field below as well?

src/main/java/org/broadinstitute/dsm/jobs/PubSubLookUp.java
yufengwng: @pegahtah as long as we don't save it and don't send it over to pepper, then it's fine with me.

@SimoneMaiwald yeah a `warn()` should be good, as long as we throw it out.

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -306,6 +318,52 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
+                        // Handle incoming message, then ack the received message.
+                        try {
+                            PubSubLookUp.processCovidTestResults(message);
+                            logger.info("Processing the message finished");
+                            consumer.ack();
+
+                        }catch(Exception ex){
+                            logger.info("about to nack the message", ex);
+                            consumer.nack();
+                            ex.printStackTrace();
```
yufengwng, line 349: We can remove this statement, unless we want double the error lines in the logs :)

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -306,6 +318,52 @@ protected void setupCustomRouting(@NonNull Config cfg) {
...
+                        }
+                    };
+
+            Subscriber subscriber = null;
+            ProjectSubscriptionName resultSubName = ProjectSubscriptionName.of(projectId, subscriptionId);
+            ExecutorProvider resultsSubExecProvider = InstantiatingExecutorProvider.newBuilder().setExecutorThreadCount(10).build();
```
yufengwng, line 355: We probably don't need 10 threads, but I'll leave that choice to you since I don't know what else is going on thread-wise.

src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -46,8 +46,8 @@
             "WHERE req.ddp_instance_id = ddp_site.ddp_instance_id AND req.kit_type_id = kt.kit_type_id) AS request " +
             "LEFT JOIN (SELECT * FROM (SELECT kit.dsm_kit_request_id, kit.dsm_kit_id, kit.kit_complete, kit.label_url_to, kit.label_url_return, kit.tracking_to_id, " +
             "kit.tracking_return_id, kit.easypost_tracking_to_url, kit.easypost_tracking_return_url, kit.easypost_to_id, kit.easypost_shipment_status, kit.scan_date, kit.label_date, kit.error, kit.message, " +
-            "kit.receive_date, kit.deactivated_date, kit.easypost_address_id_to, kit.deactivation_reason, tracking.tracking_id, kit.kit_label, kit.express, kit.needs_approval, kit.authorization, kit.denial_reason, " +
-            "kit.authorized_by FROM ddp_kit kit " +
+            "kit.receive_date, kit.deactivated_date, kit.easypost_address_id_to, kit.deactivation_reason, tracking.tracking_id, kit.kit_label, kit.express, kit.test_result, kit.needs_approval, kit.authorization, kit.denial_reason, \" +\n" +
+            "            \"kit.authorized_by FROM ddp_kit kit " +
```
yufengwng, line 50: I'm going to assume that the escaped quotes is a leftover.
```suggestion
            "kit.receive_date, kit.deactivated_date, kit.easypost_address_id_to, kit.deactivation_reason, tracking.tracking_id, kit.kit_label, kit.express, kit.test_result, kit.needs_approval, kit.authorization, kit.denial_reason, " +
            "            kit.authorized_by FROM ddp_kit kit " +
```

yufengwng (approved):
Looks fine to me.

PR #45, opened 2020-09-18
DDP-4815 cleaning up CE  ordering

zyme:
In this iteration for DDP-4815, while trying  out connectivity with Ellkay, I found some sparse data that resulted in some NPEs, so I made the  code a bit more tolerant  of missing fields. While there, I extracted some  constant fields that  I had copy-pasted previously.


src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -81,165 +90,171 @@ public  OrderResponse orderTest(Authentication auth, String participantHruid, St
...
             JsonObject data = participant.getDataAsJson();
             if (data != null) {
-                JsonObject address = data.get("address").getAsJsonObject();
-                if (address != null) {
-                    Address careEvolveAddress = toCareEvolveAddress(address);
+                if (!data.has(ADDRESS_FIELD)) {
+                    throw new CareEvolveException("No address for " + participantHruid + ".  Cannot register order.");
+                }
+                JsonObject address = data.get(ADDRESS_FIELD).getAsJsonObject();
+                Address careEvolveAddress = toCareEvolveAddress(address);
```
SimoneMaiwald, line 97: maybe add a check if address is not null?

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -81,165 +90,171 @@ public  OrderResponse orderTest(Authentication auth, String participantHruid, St
...
+                                race = "2106-3";
+                            } else if ("OTHER".equals(race)) {
                                 race = "2131-1";
+                            } else {
+                                logger.error("Could not map race " + race + " to Ellkay code; will default to other for " + baselineCovidActivity.get("guid"));
                             }
+                        } else {
+                            // if there's no value or if multiple values are selected,
+                            // use the code for "other"
+                            race = "2131-1";
```
SimoneMaiwald, line 223: could you use the string from L161

pegahtah (approved):

PR #46, opened 2020-09-21
Ups integration

pegahtah:


src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +23,213 @@
...
-                logger.info("got response back: " + response);
+        logger.info("Starting the UPS lookup job");
+        List<DDPInstance> ddpInstanceList = DDPInstance.getDDPInstanceListWithRole("ups_tracking");
+        for (DDPInstance ddpInstance : ddpInstanceList) {
+            if (ddpInstance.isHasRole()) {
+                if (ddpInstance != null) {
+                    logger.info("tracking ups ids for " + ddpInstance.getName());
+                    Map<String, Map<String, DdpKit>> ids = getResultSet(ddpInstance.getDdpInstanceId());
+                    orderRegistrar = new Covid19OrderRegistrar(DSMServer.careEvolveOrderEndpoint, DSMServer.careEvolveAccount, DSMServer.provider);
+                    Map<String, DdpKit> kits = ids.get("shipping");
```
SimoneMaiwald: potential npe if ids is null

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +23,213 @@
...
-            headers.put("Content-Type", "application/json");
-            headers.put("Accept", "application/json");
-            try {
-                UPSTrackingResponse response = DDPRequestUtil.getResponseObjectWithCustomHeader(UPSTrackingResponse.class, sendRequest, "UPS Tracking Test " + inquiryNumber, headers);
-                logger.info("got response back: " + response);
+        logger.info("Starting the UPS lookup job");
+        List<DDPInstance> ddpInstanceList = DDPInstance.getDDPInstanceListWithRole("ups_tracking");
+        for (DDPInstance ddpInstance : ddpInstanceList) {
+            if (ddpInstance.isHasRole()) {
+                if (ddpInstance != null) {
```
SimoneMaiwald: order should be different here

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +23,213 @@
...
+    public static void lookUpKit(DdpKit kit, boolean isReturn) {
+        String trackingId;
+        if (!isReturn) {
+            trackingId = kit.getTrackingToId();
+        }
+        else {
+            trackingId = kit.getTrackingReturnId();
+        }
+        String transId = NanoIdUtils.randomNanoId(
+                NanoIdUtils.DEFAULT_NUMBER_GENERATOR, "1234567890QWERTYUIOPASDFGHJKLZXCVBNM".toCharArray(), 32);
```
SimoneMaiwald: I saw that in the last PR already. where is the method to do that? =)

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +23,213 @@
...
+    private static void logError(String trackingId, UPSError[] errors) {
+        String errorString = "";
+        for (UPSError error : errors) {
+            errorString += "Got Error: " + error.getCode() + " " + error.getMessage() + " For Tracking Number " + trackingId;
+        }
+        logger.error(errorString);
+    }
+
+    private static void updateStatus(String trackingId, String oldType, UPSTrackingResponse response, boolean isReturn, DdpKit kit) {
+        UPSShipment[] shipment = response.getTrackResponse().getShipment();
```
SimoneMaiwald: potential npe

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +23,213 @@
...
+        if (shipment != null && shipment.length > 0) {
+            UPSPackage[] upsPackages = shipment[0].getUpsPackageArray();
+            if (upsPackages != null && upsPackages.length > 0) {
+                UPSPackage upsPackage = upsPackages[0];
+                UPSActivity activity = upsPackage.getActivity()[0];
+                UPSStatus status = activity.getStatus();
```
SimoneMaiwald: potential npe

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +23,213 @@
...
+                stmt.setString(2, date);
+                stmt.setString(3, trackingId);
+                int r = stmt.executeUpdate();
+                if (r != 2) {//number of subkits
+                    throw new RuntimeException("Update query for UPS tracking updated " + r + " rows!");
+                }
+                else {
+                    if (!isReturn) {
+                        //if delivered notif pepper
+                        if (statusType.equals(DELIVERY) && !(DELIVERY.equals(oldType))) {
```
SimoneMaiwald: statusType.equals(DELIVERY) change to DELIVERY.equals(statusType)

SimoneMaiwald: same for PICKUP and DELIVERY check

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +23,213 @@
...
+            String upsUpdate = statusType + " " + statusDescription;
+            try (PreparedStatement stmt = conn.prepareStatement(query)) {
+                stmt.setString(1, upsUpdate);
+                stmt.setString(2, date);
+                stmt.setString(3, trackingId);
+                int r = stmt.executeUpdate();
+                if (r != 2) {//number of subkits
+                    throw new RuntimeException("Update query for UPS tracking updated " + r + " rows!");
+                }
+                else {
```
SimoneMaiwald: I would move all that logic out of the inTransaction part and do it at the end after  if (result.resultException != null) 
 is checked

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +22,230 @@
...
+                        logger.error("delivered kitDDPNotification was null for " + kit.getExternalOrderNumber());
+                    }
+                }
 
             }
-            catch (IOException e) {
-                throw new RuntimeException("couldn't get response from ups tracking ", e);
+            else {
+                //if picked up place order
+                if (!(PICKUP.equals(oldType)) && !(IN_TRANSIT.equals(oldType)) && (PICKUP.equals(statusType) || IN_TRANSIT.equals(statusType))) {
```
zyme: Maybe factor out L207 to a method like `isBeingReturned(oldType, newType)`?

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +22,230 @@
...
+                    Instant now = Instant.now();
+                    orderRegistrar.orderTest(DSMServer.careEvolveAuth, kit.getHRUID(), kit.getKitLabel(), kit.getExternalOrderNumber(), now);
```
zyme, line 210: Is `kit.getKitLabel()` the `TBOS-` value?  From what I saw earlier in the database, this was a different label.  The `TBOS-` label is what's required here, since that ends up in `ORC-2`.

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +22,230 @@
...
+                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(SQL_SELECT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER + SELECT_BY_RETURN_NUMBER, new String[] { RECEIVED, trackingId }, 2);//todo change this to the number of subkits but for now 2 for test boston works
+                    if (kitDDPNotification != null) {
+                        logger.info("Triggering DDP for received kit with external order number: " + kit.getExternalOrderNumber());
+                        EventUtil.triggerDDP(kitDDPNotification);
+
+                    }
+                    else {
+                        logger.error("received kitDDPNotification was null for " + kit.getExternalOrderNumber());
+                    }
+                }
```
zyme, line 227: An `else` here will be helpful, just to log something like "We don't know what to do here".

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +22,230 @@
...
+        String upsUpdate = statusType + " " + statusDescription;
+        SimpleResult result = inTransaction((conn) -> {
+            SimpleResult dbVals = new SimpleResult();
+            try (PreparedStatement stmt = conn.prepareStatement(query)) {
+                stmt.setString(1, upsUpdate);
+                stmt.setString(2, date);
+                stmt.setString(3, trackingId);
+                int r = stmt.executeUpdate();
+                if (r != 2) {//number of subkits
+                    throw new RuntimeException("Update query for UPS tracking updated " + r + " rows!");
```
zyme: Maybe add some more context to this error?  `trackingId` or whatever is a handy value from `kit`.

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
pegahtah: Yes I changed that

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +22,230 @@
...
+                        logger.error("received kitDDPNotification was null for " + kit.getExternalOrderNumber());
+                    }
+                }
             }
+            logger.info("Updated status of tracking number " + trackingId + " to " + upsUpdate + " from " + oldType);
+
         }
     }
 
+    public static void testMethod() {
```
zyme: Shall we prune this method?

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -21,60 +22,230 @@
...
+            errorString += "Got Error: " + error.getCode() + " " + error.getMessage() + " For Tracking Number " + trackingId;
+        }
+        logger.error(errorString);
+    }
+
+    private static void updateStatus(String trackingId, String oldType, UPSTrackingResponse response, boolean isReturn, DdpKit kit) {
+        if (response.getTrackResponse() != null) {
+            UPSShipment[] shipment = response.getTrackResponse().getShipment();
+
+            if (shipment != null && shipment.length > 0) {
```
zyme, line 144: Do we need to error out if `shipment.length > 1`?

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
zyme: We should probably do some sort of duplicate check here--can we keep track of which kits we've ordered via `orderTest()` and only place an order for kits that don't already have an order in CE?  Placing multiple orders for the same thing in CE is to be avoided, and although Birch does some duplicate filtering of its own, we should put duplicate detection here as well.

zyme: Maybe we should store `OrderResponse.handle` in DSM, and if we see a non-blank `handle`, we don't place an order?

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
pegahtah: Yes I think we should decide on what beingReturned means for us

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
pegahtah: oops

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
zyme: I think L207 is a good start--let's move ahead with this and expand as we see other patterns.  Are there UPS mail innovation docs that describe each status somewhere?

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
pegahtah: I can keep track in dsm, would be easier to check. Does CE gives us an id o keep or I can just have a flag for the time order placed

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
pegahtah: I think the log on the next line explains that. What do you think? Because if we don't know what to do it's because the new status was irrelevant to us and the logs say it

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
pegahtah: no they are very summarized, and I can see now in logs that they are not really totally correct

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
zyme: `OrderResponse.handle` will be different even for a duplicated order, but just knowing that there's _some_ id should be sufficient.

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
zyme: Ah, got it!

zyme (approved):

PR #49, opened 2020-09-25
multiple fixes

SimoneMaiwald:
- fixing tracking return id at participant list
- pdf download fix per new lddp version
- fixing cover pdf 
- fixing sample > summary and report pages

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/route/DashboardRoute.java
+++ b/src/main/java/org/broadinstitute/dsm/route/DashboardRoute.java
@@ -32,6 +32,15 @@ public class DashboardRoute extends RequestHandler {
 
     private static final Logger logger = LoggerFactory.getLogger(DashboardRoute.class);
 
+    private static final String SQL_SELECT_KIT_REQUESTS_SENT = "SELECT kit.scan_date, realm.instance_name, request.ddp_instance_id FROM ddp_kit_request request " +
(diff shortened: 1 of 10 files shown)


src/main/java/org/broadinstitute/dsm/route/DashboardRoute.java
```diff
@@ -815,4 +792,17 @@ public void getKitRequestInformationPerMonth(@NonNull String realm, @NonNull Kit
             throw new RuntimeException("Error getting kit request information for download ", results.resultException);
         }
     }
+
+    private int getKitTypeId (@NonNull String realm, @NonNull KitType kitType) {
+        int kitTypeId = kitType.getKitId();
+        List<KitSubKits> subKits = KitSubKits.getSubKits(realm, kitType.getName());
+        if (subKits != null && !subKits.isEmpty()) {
+            //kit has sub kits (assumption: all subkits stay together and will therefore be counted as just "one" kit)
+            KitSubKits firstSubKit = subKits.get(0);
```
pegahtah, line 801: so it always returns the first subkit?

pegahtah (approved):

src/main/java/org/broadinstitute/dsm/route/DashboardRoute.java
SimoneMaiwald: it only uses the kittypeId from the first subkit to count the numbers of subkits in the database

PR #54, opened 2020-10-13
DDP-5109 CareEvolve Robustness

zyme:
DDP-5109 wraps CE ordering in a loop with `Thread.sleep()` so that if there's an issue with CE, we retry a few times before giving up.

Config file changes will need to be applied to dev and prod secret manager for `careEvolve.maxRetries` and `careEvolve.retryWaitSeconds`.  Suggest we do 3 attempts with a 5s pause between each.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/DSMServer.java
+++ b/src/main/java/org/broadinstitute/dsm/DSMServer.java
@@ -111,6 +111,8 @@ public class DSMServer extends BasicServer {
     public static String careEvolveSubscriberKey;
     public static String careEvolveServiceKey;
     public static String careEvolveOrderEndpoint;
+    public static int careEvolveMaxRetries;
+    public static int careEvolveRetyWaitSeconds;
     public static String careEvolveAccount;
     public static Authentication careEvolveAuth;
     public static Provider provider;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -586,6 +588,17 @@ private void setupJobs(@NonNull Config cfg, @NonNull KitUtil kitUtil,
...
                 careEvolveOrderEndpoint = cfg.getString(ApplicationConfigConstants.CARE_EVOLVE_ORDER_ENDPOINT);
+                if (cfg.hasPath(ApplicationConfigConstants.CARE_EVOLVE_MAX_RETRIES)) {
+                    careEvolveMaxRetries = cfg.getInt(ApplicationConfigConstants.CARE_EVOLVE_MAX_RETRIES);
+                } else {
+                    careEvolveMaxRetries = 5;
+                }
+                if (cfg.hasPath(ApplicationConfigConstants.CARE_EVOLVE_RETRY_WAIT_SECONDS)) {
+                    careEvolveRetyWaitSeconds = cfg.getInt(ApplicationConfigConstants.CARE_EVOLVE_RETRY_WAIT_SECONDS);
+                } else {
+                    careEvolveRetyWaitSeconds = 10;
```
zyme, line 599: Hardcoded defaults in case someone forgets to update a config file...

pegahtah (approved):

PR #61, opened 2020-11-19
DDP-5323 Use in transit and delivered status to trigger events

zyme:
DDP-5323 is about triggering symptom surveys based on `delivered` _or_ `in transit` states due to recent experience where UPS does not post delivered events, leaving participants in a state where they have a kit but they don't have a symptom status.

@pegahtah @SimoneMaiwald please let me know if this is the right branch for a hotfix.  What is the best way to test this?

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
@@ -13,7 +13,6 @@ import org.broadinstitute.dsm.util.DDPRequestUtil;
 import org.broadinstitute.dsm.util.EventUtil;
 import org.broadinstitute.dsm.util.KitUtil;
 import org.broadinstitute.dsm.util.NanoIdUtil;
-import org.broadinstitute.dsm.util.externalShipper.GBFRequestUtil;
 import org.quartz.Job;
 import org.quartz.JobExecutionContext;
 import org.quartz.JobExecutionException;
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -235,6 +232,15 @@ else if (statusType.equals(DELIVERY) && !(DELIVERY.equals(oldType))) {
         }
     }
 
+    /**
+     * Determines whether or not a trigger should be sent to
+     * study-server to respond to kit being sent to participant
+     */
+    private static boolean shouldTriggerKitDeliveryEvent(String currentStatus, String previousStatus) {
```
zyme, line 239: Intent here is that if the kit is newly in the `D` or `I` category, an event should be sent.  But state changes from `I` to `D` should not trigger an update.  Only a change from something other than `D` or `I` into `D` or `I` state should trigger the event.

yufengwng: If kit is in `in_transit`, how long does it typically take for it to be physically delivered in the real world? The "delivered" event will trigger the symptom surveys, one of which is the "longitudinal" covid survey. That one has an expiration time of 7 days, after which it becomes read-only. I imagine it might only take 1 day or so for that state transition, so users should have plenty of time?

yufengwng (commented):
Looks good to me.

PR #63, opened 2020-11-21
DDP-5323 fixing premature read-only of tbos symptoms activity

zyme:
DDP-5323 Fixing bug that was prematurely marking a kit as returned to broad and therefore locking the longitudinal activity before the kit was delivered to the participant.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
@@ -192,7 +192,7 @@ public class UPSTrackingJob implements Job {
         }
         else {
             if (!isReturn) {
-                if (shouldTriggerKitDeliveryEvent(statusType, oldType)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -213,7 +213,7 @@ private static void updateTrackingInfo(String statusType, String oldType, String
                     logger.info("Placed CE order for kit with external order number " + kit.getExternalOrderNumber());
                     kit.changeCEOrdered(true);
                 }
-                else if (shouldTriggerKitDeliveryEvent(statusType, oldType)) {
```
zyme, line 216: Previous update used the same logic for kits going to participant and kits coming back.  We need to different criteria: one for the outbound kit being in transit (or delivered) and another one for the kit being delivered back at broad (but not just in transit).

yufengwng (commented):
LGTM

PR #64, opened 2020-11-22
DDP-5345 Use UPS history for collection date

zyme:
DDP-5345 is about ignoring participant collection date and instead using UPS shipping records to overcome inconsistently entered data until we can resolve ambiguities in the kit/activity relationship that is currently inferred based on time.

A few things to note in this PR:
* Collection date will always be the earliest inbound UPS date for the kit, regardless of patient-reported collection time.
* Isolated UPS tracking lookup to a separate class
* Added methods to UPSPackage and UPSStatus to simplify how we look for in transit-ish statuses
* Removed code related to looking up most recent symptom activity\
* The fallback of `Instant.now()` for collection date has been removed


src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -241,13 +192,20 @@ public  OrderResponse orderTest(Authentication auth, String participantHruid, St
                             }
                         }
                     }
+                }
 
+                if (StringUtils.isBlank(firstName)) {
```
zyme, line 197: Added a few more checks here to prevent downstream errors in LKTransfer/Birch.

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -284,27 +242,8 @@ public  OrderResponse orderTest(Authentication auth, String participantHruid, St
...
             }
         } else {
-            throw new CareEvolveException("No particiant data found for " + participantHruid + ".  Cannot register order.");
-        }
-    }
-
-    /**
-     * Returns the latest kit-related activity
-     */
-    private JsonObject getLatestKitActivity(JsonArray activities) {
```
zyme, line 294: Don't need this any more, so I removed it.

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -195,25 +188,26 @@ private static void updateTrackingInfo(String statusType, String oldType, String
...
                     else {
                         logger.error("delivered kitDDPNotification was null for " + kit.getExternalOrderNumber());
                     }
                 }
 
             }
-            else {
-                //if picked up place order
-                if (shouldMakeCEOrder(statusType) && !kit.isCEOrdered()) {
-                    Instant now = Instant.now();
```
zyme, line 211: This `Instant.now()` under the right circumstances is usually fairly close to the shipping time, but it's not precise.  Depending on how much time has elapsed between this method's awareness of a kit coming back and the actual dates, the collection date used via `Instant.now()` can be considerably far from reality.

src/main/java/org/broadinstitute/dsm/model/ups/UPSActivity.java
```diff
@@ -8,4 +12,43 @@
     UPSStatus status;
     String date;
     String time;
+
+    public UPSActivity() {}
```
zyme, line 16: Adding explicit no-arg constructors here since this looks like it's used in various reflection based serialization/deserialization tools.

src/main/java/org/broadinstitute/dsm/model/ups/UPSActivity.java
```diff
@@ -8,4 +12,43 @@
...
+
+    public UPSActivity(UPSStatus status, String date, String time) {
+        this.status = status;
+        this.date = date;
+        this.time = time;
+    }
+
+    public String getDateTimeString() {
+        return date + " " + time;
```
zyme, line 25: Preserving the current date format in case any db tools rely on it.

src/main/java/org/broadinstitute/dsm/shipping/UPSTracker.java
```diff
@@ -0,0 +1,54 @@
...
+                      String password,
+                      String accessKey) {
+
+        this.baseEndpoint = baseEndpoint;
+        this.username = username;
+        this.password = password;
+        this.accessKey = accessKey;
+    }
+
+    public UPSTrackingResponse lookupTrackingInfo(String trackingId) {
```
zyme, line 33: Factored this out into its own class so that we can inspect various kit histories during development.

src/main/java/org/broadinstitute/dsm/model/ups/UPSActivity.java
```diff
@@ -8,4 +12,43 @@
...
+    }
+
+    public Instant getInstant() {
+        Instant eventTime = null;
+        String dateTime = getDateTimeString();
+        if (dateTime != null) {
+            try {
+                eventTime = new SimpleDateFormat("yyyyMMdd kkmmss").parse(dateTime).toInstant();
```
yufengwng: 1) What timezone are the UPS timestamps in? Are they in UTC? I believe `SimpleDateFormat` assumes _local_ timezone. Is DSM always running in UTC timezone or EST timezone? You might want to call `setTimeZone()` to explicitly set UTC timezone.

2) I recommend using the newer class `DateTimeFormatter`.

3) Curious why the hour pattern is `kk` instead of `HH`? Is that a peculiarity of the UPS timestamps?

yufengwng (commented):
Looks good as far as I can tell.

src/main/java/org/broadinstitute/dsm/model/ups/UPSActivity.java
zyme: According to the UPS docs, the time zone is local time for the place where the event happened.  So figuring this out would need a bit more work--looking at the `location` object and figuring out the zone.  From the docs, I can't see an obvious place to get the zone from the API.

Hour should be `HH`!  Thanks for that.

zyme: @yufengwng one more round?  Given the time sensitive (hahaha pun) nature of this and the geographic constraints of testboston, I went for `America/New_York` and documented things accordingy.

yufengwng (approved):
LGTM

PR #65, opened 2020-11-30
DDP-5353 gbf start time

zyme:
DDP-5353 moves GBF's start time to a fixed date.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
@@ -33,13 +33,8 @@ public class ExternalShipperJob implements Job {
                 ArrayList<KitRequest> kitRequests = shipper.getKitRequestsNotDone(kitType.getInstanceId());
                 shipper.orderStatus(kitRequests);
                 if (kitRequests != null && !kitRequests.isEmpty()) { // only if there are kits which are not yet having kit_label set
-                    JobDataMap dataMap = context.getJobDetail().getJobDataMap();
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
```diff
@@ -34,16 +34,9 @@ public void execute(JobExecutionContext context) {
                     JobDataMap dataMap = context.getJobDetail().getJobDataMap();
                     String additionalCronExpression = (String) dataMap.get(DSMServer.ADDITIONAL_CRON_EXPRESSION);
                     if (CronExpression.isValidExpression(additionalCronExpression)) {
```
SimoneMaiwald: can also be removed if we don't use the additional cron expression

src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
```diff
@@ -33,15 +34,9 @@ public void execute(JobExecutionContext context) {
...
-                        if (cron.isSatisfiedBy(context.getFireTime())) {
-                            long lastRun = DBUtil.getBookmark(DBConstants.GBF_CONFIRMATION);
-                            long now = System.currentTimeMillis();
-                            shipper.orderConfirmation(kitRequests, lastRun, now);
-                        }
-                    }
-                    else {
-                        throw new RuntimeException("Additional cron expression is not a valid cron expression");
+                        long now = System.currentTimeMillis();
+                        long fixedStartTime = new SimpleDateFormat("yyyy-MM-dd").parse("2020-08-01").getTime();
```
zyme: Actually this is probably a bit tooooo long.  Will change this to be today - 30 days...

PR #66, opened 2020-12-02
DDP-5375 GBF/UPS fault tolerance

zyme:
DDP-5375 came about because of [this accessioning issue](https://labopsjira.broadinstitute.org/browse/BQMS-13458).

The root cause is actually Ellkay Transfer not properly parsing an `AE` ack coming back from Birch.  However, while perusing the code, there are a few things here that should result in better fault tolerance.  See inline comments for details.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/db/DdpKit.java
+++ b/src/main/java/org/broadinstitute/dsm/db/DdpKit.java
@@ -1,13 +1,12 @@
 package org.broadinstitute.dsm.db;
 
 import lombok.Data;
-import org.broadinstitute.ddp.db.SimpleResult;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.sql.Connection;
 import java.sql.PreparedStatement;
-
-import static org.broadinstitute.ddp.db.TransactionWrapper.inTransaction;
+import java.sql.SQLException;
 
 @Data
 public class DdpKit {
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/dsm/db/DdpKit.java
```diff
@@ -47,30 +46,20 @@ public DdpKit(String DsmKitRequestId, String kitLabel, String trackingToId, Stri
         this.CEOrdered = CEOrdered;
     }
 
-    public void changeCEOrdered(boolean orderStatus) {
+    public void changeCEOrdered(Connection conn, boolean orderStatus) {
```
zyme, line 49: Passing through the connection so that all operations can go through on a single transaction.

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -164,65 +165,59 @@ private static void updateTrackingInfo(String statusType,
                                            DdpKit kit,
                                            Instant earliestInTransitTime) {
         String upsUpdate = statusType + " " + statusDescription;
-        SimpleResult result = inTransaction((conn) -> {
```
zyme, line 167: Because this runs in its own transaction, if there is an error outside this block, the ups status might be updated--potentially to `D` or `I`--which will then result in subsequent calls to `shouldTriggerEventForKitOnItsWayToParticipant()` or `shouldTriggerEventForReturnKitDelivery()` returning `false` when they should return `true` because a subsequent run of the job will find a new value for `oldType`.  So

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -164,65 +165,59 @@ private static void updateTrackingInfo(String statusType,
...
-                        EventUtil.triggerDDP(kitDDPNotification);
-                    }
-                    else {
-                        logger.error("delivered kitDDPNotification was null for " + kit.getExternalOrderNumber());
-                    }
+        } else { // this is a return
+            if (earliestInTransitTime != null && !kit.isCEOrdered()) {
+                // if we have the first date of an inbound event, create an order in CE
+                // using the earliest date of inbound event
+                OrderResponse orderResponse = orderRegistrar.orderTest(DSMServer.careEvolveAuth, kit.getHRUID(), kit.getKitLabel(), kit.getExternalOrderNumber(), earliestInTransitTime);
```
zyme, line 182: While here, I figured we should check the `orderResponse` in case there's an error.

src/main/java/org/broadinstitute/dsm/jobs/UPSTrackingJob.java
```diff
@@ -164,65 +165,59 @@ private static void updateTrackingInfo(String statusType,
...
-                if (shouldTriggerEventForReturnKitDelivery(statusType, oldType)) {
-                    KitUtil.setKitReceived(kit.getKitLabel());
-                    logger.info("RECEIVED: " + trackingId);
-                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(SQL_SELECT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER + SELECT_BY_RETURN_NUMBER, new String[] { RECEIVED, trackingId }, 2);//todo change this to the number of subkits but for now 2 for test boston works
-                    if (kitDDPNotification != null) {
-                        logger.info("Triggering DDP for received kit with external order number: " + kit.getExternalOrderNumber());
-                        EventUtil.triggerDDP(kitDDPNotification);
+                logger.info("Placed CE order for kit " + kit.getKitLabel() + " with external order number " + kit.getExternalOrderNumber() + " and CE ack" + orderResponse.getHandle());
 
+                TransactionWrapper.inTransaction(conn -> {
```
zyme, line 189: The new approach here is to bundle some db updates into a single transaction here if CE ordering completes.  Birch is already doing de-dup, so if a subsequent job incorrectly attempts to place a redundant order, Birch will ignore it.

PR #68, opened 2020-12-09
DDP-5404 added detailed logging for order response and expanded retry…

zyme:
DDP-5404 added detailed logging for order response and expanded retry wait and number of retries.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
@@ -145,29 +145,25 @@ public class GBFRequestUtil implements ExternalShipper {
             }
             if (!orders.getOrders().isEmpty()) {
                 String orderXml = GBFRequestUtil.orderXmlToString(Orders.class, orders);
-
-                //                logger.info("orderXML: " + orderXml);
                 boolean test = DSMServer.isTest(getExternalShipperName());//true for dev in `not-secret.conf`
                 JSONObject payload = new JSONObject().put("orderXml", orderXml).put("test", test);
                 String sendRequest = DSMServer.getBaseUrl(getExternalShipperName()) + ORDER_ENDPOINT;
                 String apiKey = DSMServer.getApiKey(getExternalShipperName());
                 Response gbfResponse = null;
-                int totalAttempts = 2 + additionalAttempts;
                 Exception ex = null;
                 try {
-                    for (int i = 1; i <= totalAttempts; i++) {
+                    for (int i = 1; i <= MAX_ORDER_RETRIES; i++) {
                         //if this isn't the first attempt let's wait a little before retrying...
                         if (i > 1) {
-                            logger.info("Sleeping before request retry for " + sleepInMs + " ms.");
-                            Thread.sleep(sleepInMs);
-                            logger.info("Sleeping done.");
+                            logger.info("Sleeping before request retry for " + GBF_SLEEP_BETWEEN_RETRIES + " ms.");
+                            Thread.sleep(GBF_SLEEP_BETWEEN_RETRIES);
                         }
                         try {
                             gbfResponse = executePost(Response.class, sendRequest, payload.toString(), apiKey);
                             break;
                         }
                         catch (Exception newEx) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -82,8 +82,8 @@
 
     public final String EXTERNAL_SHIPPER_NAME = "gbf";
 
-    private static int additionalAttempts = 1;
-    private static int sleepInMs = 500;
+    private static final int MAX_ORDER_RETRIES = 5;
```
zyme, line 85: Since the intent is for these to be constants, they should be final.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -82,8 +82,8 @@
...
+    private static final int GBF_SLEEP_BETWEEN_RETRIES = 7_000;
```
zyme, line 86: 500ms is probably too short to overcome a brief network issue or GBF outage.  I upped this to 5 retries and 7 seconds to be on the safe side.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -145,29 +145,27 @@ else if (address != null) {
...
-                int totalAttempts = 2 + additionalAttempts;
                 Exception ex = null;
                 try {
-                    for (int i = 1; i <= totalAttempts; i++) {
+                    for (int i = 1; i <= MAX_ORDER_RETRIES; i++) {
                         //if this isn't the first attempt let's wait a little before retrying...
                         if (i > 1) {
-                            logger.info("Sleeping before request retry for " + sleepInMs + " ms.");
-                            Thread.sleep(sleepInMs);
-                            logger.info("Sleeping done.");
```
zyme, line 163: Because the next lines themselves issue log statements, the "sleeping done" is perhaps a bit redundant.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -145,29 +145,27 @@ else if (address != null) {
...
                         }
                         try {
+                            logger.info("Attempting to order {} kits via {}", orders.getOrders().size(), sendRequest);
                             gbfResponse = executePost(Response.class, sendRequest, payload.toString(), apiKey);
+                            logger.info("Completed order for {} kits via {}", orders.getOrders().size(), sendRequest);
                             break;
                         }
                         catch (Exception newEx) {
-                            logger.warn("Send request failed (attempt #" + i + " of " + totalAttempts + "): ", newEx);
+                            logger.error("Send request failed (attempt #" + i + " of " + MAX_ORDER_RETRIES + ")", newEx);
```
zyme: Seems worthy of an error and an alert.  The reader of the message will hopefully be reassured by the attempt counts and the fact that subsequent messages may clear up the hopefully transient issue.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -176,11 +174,13 @@ else if (address != null) {
                     throw new RuntimeException("Unable to send requests.", ex);
                 }
                 if (gbfResponse != null && gbfResponse.isSuccess()) {
-                    logger.info("Ordered kits");
+                    for (Status kitStatus : gbfResponse.getStatuses()) {
+                        logger.info("Order status from order endpoint for " + kitStatus.getOrderNumber() + " is " + kitStatus.getOrderStatus());
```
zyme, line 178: TBOS team was asking for some specific numbers about when various orders were placed, so explicitly logging the order number here will make it easier to answer questions.  Prior to this, one would have to grep the logs looking for above statements, and then re-grep to find a correlation in time between that log statement and an indication of success.  But one would also need to look at other log messages to ensure that the order wasn't filtered out due to a bad address.

PR #77, opened 2020-12-28
prod_fix_12282020_ddpParticipantDropped

pegahtah:
removed ES calls when having confirmation xmls

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
@@ -354,7 +354,7 @@ public class GBFRequestUtil implements ExternalShipper {
                 try (ResultSet rs = stmt.executeQuery()) {
                     while (rs.next()) {
                         String ddpParticipantId = rs.getString(DBConstants.DDP_PARTICIPANT_ID);
-                        logger.info("querying ES for ddpParticipantId: "+ddpParticipantId);
+//                        logger.info("querying ES for ddpParticipantId: "+ddpParticipantId);
                         if (StringUtils.isNotBlank(ddpParticipantId)) {
 //                            Map participantsESData = getParticipant(ddpInstance, ddpParticipantId);
 //                            if (participantsESData != null && !participantsESData.isEmpty()) {
--- a/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
@@ -343,7 +343,7 @@ public class GBFRequestUtil implements ExternalShipper {
         return getKitRequestsNotDone(instanceId,SQL_SELECT_EXTERNAL_KIT_NOT_DONE );
     }
 
-    public  ArrayList<KitRequest> getKitRequestsNotDone(int instanceId, String query) {
+    public  ArrayList<KitRequest>   getKitRequestsNotDone(int instanceId, String query) {
         DDPInstance ddpInstance = DDPInstance.getDDPInstanceById(instanceId);
 
         ArrayList<KitRequest> kitRequests = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -356,21 +356,21 @@ public void orderCancellation(ArrayList<KitRequest> kitRequests) throws Exceptio
                         String ddpParticipantId = rs.getString(DBConstants.DDP_PARTICIPANT_ID);
                         logger.info("querying ES for ddpParticipantId: "+ddpParticipantId);
```
SimoneMaiwald: please remove that logging as well, otherwise we will wonder why we are not seeing any logging for the ES querying

pegahtah: ok

src/test/java/org/broadinstitute/dsm/GBFTest.java
```diff
@@ -178,8 +178,9 @@ public void inputConfirmationGBFKit() throws Exception {
             GBFRequestUtil gbf = new GBFRequestUtil();
                 String query = "SELECT * " +
                         "FROM ddp_kit_request req  " +
-                        "LEFT JOIN ddp_kit kit ON  " +
-                        "(req.dsm_kit_request_id = kit.dsm_kit_request_id) " +
+                        "LEFT JOIN ddp_kit kit ON (req.dsm_kit_request_id = kit.dsm_kit_request_id)  " +
+                        "LEFT JOIN ddp_participant_exit ex ON (ex.ddp_instance_id = req.ddp_instance_id AND ex.ddp_participant_id = req.ddp_participant_id)   " +
```
SimoneMaiwald: you don't need to join with `ddp_participant_exit` table if it is not used in the query

src/test/java/org/broadinstitute/dsm/GBFTest.java
```diff
@@ -168,25 +169,30 @@ public void inputConfirmationGBFKit() throws Exception {
...
-                        "(req.dsm_kit_request_id = kit.dsm_kit_request_id) " +
-                        "WHERE " +
-                        "req.ddp_instance_id = ?  " +
-                        "AND external_order_status = 'SHIPPED' " +
-                        "AND external_response is null " +
-                        "ORDER BY external_order_date ASC ";
+            String query = "SELECT * " +
+                    "FROM ddp_kit_request req  " +
+                    "LEFT JOIN ddp_kit kit ON (req.dsm_kit_request_id = kit.dsm_kit_request_id)  " +
+                    "LEFT JOIN ddp_participant_exit ex ON (ex.ddp_instance_id = req.ddp_instance_id AND ex.ddp_participant_id = req.ddp_participant_id)   " +
```
SimoneMaiwald: ddp_participant_exit can be removed if not used

pegahtah: ok

src/test/java/org/broadinstitute/dsm/GBFTest.java
```diff
@@ -168,25 +169,30 @@ public void inputConfirmationGBFKit() throws Exception {
...
+                    "req.ddp_instance_id = ?  " +
+                    "AND external_order_status = 'SHIPPED' " +
+                    "AND external_response is null " +
+                    "ORDER BY external_order_date ASC ";
 
             ArrayList<KitRequest> kitRequests = gbf.getKitRequestsNotDone(Integer.parseInt(DDPInstance.getDDPInstance("testboston").getDdpInstanceId()), query);
+            HashMap<String, KitRequest> kits = new HashMap<>();
+            for (KitRequest kit : kitRequests) {
+                kits.put(kit.getExternalOrderNumber(), kit);
+            }
```
pegahtah, line 194: This is the map @SimoneMaiwald

PR #78, opened 2020-12-29
DDP-5479 refactored GBF confirmation

SimoneMaiwald:
DDP-5479

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
@@ -27,6 +27,7 @@ public class ExternalShipperJob implements Job {
                 ExternalShipper shipper = (ExternalShipper) Class.forName(DSMServer.getClassName(kitType.getExternalShipper())).newInstance();//GBFRequestUtil
                 ArrayList<KitRequest> kitRequests = shipper.getKitRequestsNotDone(kitType.getInstanceId());
                 shipper.orderStatus(kitRequests);
+                kitRequests.clear();
                 Instant now = Instant.now();
                 Instant dynamicStartTime = now.minus(5, ChronoUnit.DAYS);
                 shipper.orderConfirmation(dynamicStartTime.toEpochMilli(), now.toEpochMilli());
--- a/src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
@@ -3,17 +3,12 @@ package org.broadinstitute.dsm.jobs;
 import org.broadinstitute.dsm.DSMServer;
 import org.broadinstitute.dsm.model.KitRequest;
 import org.broadinstitute.dsm.model.KitType;
-import org.broadinstitute.dsm.statics.DBConstants;
-import org.broadinstitute.dsm.util.DBUtil;
 import org.broadinstitute.dsm.util.externalShipper.ExternalShipper;
-import org.quartz.CronExpression;
 import org.quartz.Job;
-import org.quartz.JobDataMap;
 import org.quartz.JobExecutionContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.text.SimpleDateFormat;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
@@ -32,12 +27,9 @@ public class ExternalShipperJob implements Job {
                 ExternalShipper shipper = (ExternalShipper) Class.forName(DSMServer.getClassName(kitType.getExternalShipper())).newInstance();//GBFRequestUtil
                 ArrayList<KitRequest> kitRequests = shipper.getKitRequestsNotDone(kitType.getInstanceId());
                 shipper.orderStatus(kitRequests);
-                if (kitRequests != null && !kitRequests.isEmpty()) { // only if there are kits which are not yet having kit_label set
-                    logger.info("Working on " + kitRequests.size() + " incomplete external kits");
-                    Instant now = Instant.now();
-                    Instant dynamicStartTime = now.minus(5, ChronoUnit.DAYS);
-                    shipper.orderConfirmation(kitRequests, dynamicStartTime.toEpochMilli(), now.toEpochMilli());
-                }
+                Instant now = Instant.now();
+                Instant dynamicStartTime = now.minus(5, ChronoUnit.DAYS);
+                shipper.orderConfirmation(dynamicStartTime.toEpochMilli(), now.toEpochMilli());
             }
             catch (Exception e) {
                 throw new RuntimeException("Failed to get status and/or confirmation ", e);
--- a/src/main/java/org/broadinstitute/dsm/model/KitRequestExternal.java
+++ b/src/main/java/org/broadinstitute/dsm/model/KitRequestExternal.java
@@ -1,10 +1,12 @@
 package org.broadinstitute.dsm.model;
 
+import lombok.NonNull;
 import org.broadinstitute.ddp.db.SimpleResult;
 import org.broadinstitute.dsm.model.ddp.DDPParticipant;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.sql.Connection;
 import java.sql.PreparedStatement;
 
 import static org.broadinstitute.ddp.db.TransactionWrapper.inTransaction;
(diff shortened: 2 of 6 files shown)


src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
```diff
@@ -32,12 +27,9 @@ public void execute(JobExecutionContext context) {
                 ExternalShipper shipper = (ExternalShipper) Class.forName(DSMServer.getClassName(kitType.getExternalShipper())).newInstance();//GBFRequestUtil
                 ArrayList<KitRequest> kitRequests = shipper.getKitRequestsNotDone(kitType.getInstanceId());
                 shipper.orderStatus(kitRequests);
-                if (kitRequests != null && !kitRequests.isEmpty()) { // only if there are kits which are not yet having kit_label set
-                    logger.info("Working on " + kitRequests.size() + " incomplete external kits");
-                    Instant now = Instant.now();
-                    Instant dynamicStartTime = now.minus(5, ChronoUnit.DAYS);
-                    shipper.orderConfirmation(kitRequests, dynamicStartTime.toEpochMilli(), now.toEpochMilli());
-                }
+                Instant now = Instant.now();
```
SimoneMaiwald, line 31: removed kitRequests from orderConfirmation because it seemed obsolete

src/main/java/org/broadinstitute/dsm/model/KitRequestExternal.java
```diff
@@ -53,55 +55,39 @@ public static void updateKitRequest(String externalOrderStatus, long externalOrd
...
-            SimpleResult dbVals = new SimpleResult();
-            try (PreparedStatement stmt = conn.prepareStatement(SQL_UPDATE_KIT_REQUEST_EXTERNAL_SHIPPER_RESPONSE)) {
-                stmt.setString(1, externalResponse);
-                stmt.setString(2, dsmKitRequestId);
-
-                int result = stmt.executeUpdate();
-                if (result > 1) {
-                    throw new RuntimeException("Error updating kit request w/ dsm_kit_request_id " + dsmKitRequestId + " it was updating " + result + " rows");
-                }
+    public static void updateKitRequestResponse(@NonNull Connection conn, String externalResponse, String dsmKitRequestId) {
```
SimoneMaiwald, line 58: using same connection now to update kit request and kit information of one external order

src/main/java/org/broadinstitute/dsm/model/KitRequestExternal.java
```diff
@@ -53,55 +55,39 @@ public static void updateKitRequest(String externalOrderStatus, long externalOrd
...
-            logger.error("Error updating kit request w/ dsm_kit_request_id " + dsmKitRequestId, results.resultException);
+        }
+        catch (Exception e) {
+            logger.error("Error updating kit request w/ dsm_kit_request_id " + dsmKitRequestId, e);
         }
     }
 
     // update kit request with response of external shipper
-    public static void updateKitRequestResponse(String trackingIdTo, String trackingIdReturn, String kitLabel, long sentDate,
+    public static void updateKitRequestResponse(@NonNull Connection conn, String trackingIdTo, String trackingIdReturn, String kitLabel, long sentDate,
```
SimoneMaiwald, line 74: using same connection

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -269,45 +270,44 @@ else if (status.getOrderStatus().contains("CANCELLED") && (StringUtils.isBlank(k
...
+        if (confirmation.getItem() != null && StringUtils.isNotBlank(externalResponse)) {
             Item item = confirmation.getItem();
-            List<KitRequest> kitRequestsResult = getKitRequest(kitRequests, confirmation.getOrderNumber(), item.getItemNumber(), 1);
-            int counter = 0;
-            if (kitRequestsResult != null) {
-                for (KitRequest kitRequest : kitRequestsResult) {
-                    String kitLabel = item.getTubeSerial();
-                    if (counter > 0) {
-                        kitLabel += "_" + counter;
+            inTransaction((conn) -> {
```
SimoneMaiwald, line 278: making now 1 connection for request and kit information update

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -269,45 +270,44 @@ else if (status.getOrderStatus().contains("CANCELLED") && (StringUtils.isBlank(k
...
+                List<String> dsmKitRequestIds = getDSMKitRequestId(conn, confirmation.getOrderNumber()); //dsm_kit_request_id
```
SimoneMaiwald: because we are getting here the list of dsm_kit_request_ids we don't need to give the list in from before

SimoneMaiwald: we could maybe add a check, if the kit already has confirmation and all other informations, then don't update it again… but seeing gbf changing the xml I think we should leave it like that for now…

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -322,12 +322,13 @@ public void orderConfirmation(ArrayList<KitRequest> kitRequests, long startDate,
...
                     for (ShippingConfirmation confirmation : confirmationList) {
                         try {
-                            processingSingleConfirmation(gbfResponse, kitRequests, confirmation);
+                            processingSingleConfirmation(gbfResponse, confirmation);
                         } catch (Exception e) {
                             logger.error("Could not process confirmation for " + confirmation.getOrderNumber(), e);
                         }
                     }
-                    DBUtil.updateBookmark(endDate, DBConstants.GBF_CONFIRMATION);
+                    DBUtil.updateBookmark(endDate, DBConstants.GBF_CONFIRMATION); //TODO can be removed because not used anymore
```
SimoneMaiwald, line 330: I left it in for now in case we want to change it back to using the bookmark as a start point for the confirmation call

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
zyme: Is `item` always non-null?  Before, we had a check for this, so I'm wondering if something might NPE without it.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
zyme: I agree!

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -389,6 +390,22 @@ public void orderCancellation(ArrayList<KitRequest> kitRequests) throws Exceptio
         return kitRequests;
     }
 
+    public static  List<String> getDSMKitRequestId(@NonNull Connection conn, String externalOrderNumber) {
+        List<String> dsmKitRequestIds = new ArrayList<>();
+        try (PreparedStatement stmt = conn.prepareStatement(SQL_SELECT_KIT_REQUEST_BY_EXTERNAL_ORDER_NUMBER)) {
+            stmt.setString(1, externalOrderNumber);
+            try (ResultSet rs = stmt.executeQuery()) {
+                while (rs.next()) {
```
zyme, line 398: Should we do a check for at most (or exactly) one row in the result set?

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -389,6 +390,22 @@ public void orderCancellation(ArrayList<KitRequest> kitRequests) throws Exceptio
...
+                    dsmKitRequestIds.add(rs.getString(DBConstants.DSM_KIT_REQUEST_ID));
+                }
+            }
+        }
+        catch (SQLException ex) {
+            throw new RuntimeException("Error looking up kit requests  ", ex);
```
zyme: Maybe add `externalOrderNumber` to this exception text so we know a bit more context when debugging?

src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
zyme: Nice!  That will speed things up and free a bunch of memory too.

zyme: Never mind--now I see that we need it for other things anyway...

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
zyme: Ah, I see the check up above--maybe move `Item item = confirmation.getItem();` before L277 and then change `confirmation.getItem() != null` to `item != null`?

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -389,6 +390,22 @@ public void orderCancellation(ArrayList<KitRequest> kitRequests) throws Exceptio
         return kitRequests;
     }
 
+    public static  List<String> getDSMKitRequestId(@NonNull Connection conn, String externalOrderNumber) {
```
zyme: Maybe rename to plural `getDSMKitRequestIds`

src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
zyme: Maybe add `kitRequests.clear()` after L29 so that we can free this memory before the long running `orderConfirmation()` method, since we don't need the object any more.

src/main/java/org/broadinstitute/dsm/model/KitRequestExternal.java
zyme: Yay transactional!

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
zyme: edit: not relevant, going to delete...

zyme (commented):
Looks good to me overall--THANK YOU!  Just a few small suggestions!

PR #80, opened 2020-12-30
DDP-5479 Stream GBF Updates

zyme:
For DDP-5479, Instead of extracting long lists of `KitRequests` and passing them around--which is contributing to memory pressure--these changes stream things a bit more so that as we work through the result set of pending kits, we update its status from GBF as we go.

While making these changes, I found another latent bug that can result in inconsistent transactions.  I fixed this by passing around `Connection`s, which rippled through a number of classes unrelated to the driving changes above.

I also suspect that some of our larger log statements--which pull fairly large objects into a `String`--may be contributing to our problems, so I removed those.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
@@ -25,9 +25,7 @@ public class ExternalShipperJob implements Job {
             try {
                 logger.info("Starting the external shipper job");
                 ExternalShipper shipper = (ExternalShipper) Class.forName(DSMServer.getClassName(kitType.getExternalShipper())).newInstance();//GBFRequestUtil
-                ArrayList<KitRequest> kitRequests = shipper.getKitRequestsNotDone(kitType.getInstanceId());
-                shipper.orderStatus(kitRequests);
-                kitRequests.clear();
+                shipper.updateOrderStatusForPendingKitRequests(kitType.getInstanceId());
                 Instant now = Instant.now();
                 Instant dynamicStartTime = now.minus(5, ChronoUnit.DAYS);
                 shipper.orderConfirmation(dynamicStartTime.toEpochMilli(), now.toEpochMilli());
--- a/src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
@@ -181,8 +181,7 @@ public class TestBostonUPSTrackingJob implements Job {
                                            DdpKit kit,
                                            Instant earliestInTransitTime) {
         String upsUpdate = statusType + " " + statusDescription;
-        SimpleResult result = inTransaction((conn) -> {
-            SimpleResult dbVals = new SimpleResult();
+        inTransaction((conn) -> {
             try (PreparedStatement stmt = conn.prepareStatement(query)) {
                 stmt.setString(1, upsUpdate);
                 stmt.setString(2, date);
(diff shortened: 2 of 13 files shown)


src/main/java/org/broadinstitute/dsm/jobs/ExternalShipperJob.java
```diff
@@ -25,9 +25,7 @@ public void execute(JobExecutionContext context) {
             try {
                 logger.info("Starting the external shipper job");
                 ExternalShipper shipper = (ExternalShipper) Class.forName(DSMServer.getClassName(kitType.getExternalShipper())).newInstance();//GBFRequestUtil
-                ArrayList<KitRequest> kitRequests = shipper.getKitRequestsNotDone(kitType.getInstanceId());
-                shipper.orderStatus(kitRequests);
-                kitRequests.clear();
+                shipper.updateOrderStatusForPendingKitRequests(kitType.getInstanceId());
```
zyme, line 28: Instead of creating a large object here and passing it around, this changes things to streaming.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -191,56 +190,52 @@ private static void updateTrackingInfo(String statusType,
...
-                if (shouldTriggerEventForKitOnItsWayToParticipant(statusType, oldType)) {
-                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(SQL_SELECT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER + SELECT_BY_TRACKING_NUMBER, new String[] { DELIVERED, trackingId }, 2);//todo change this to the number of subkits but for now 2 for test boston works
-                    if (kitDDPNotification != null) {
-                        logger.info("Triggering DDP for kit going to participant with external order number: " + kit.getExternalOrderNumber());
-                        EventUtil.triggerDDP(kitDDPNotification);
-                    }
-                    else {
-                        logger.error("delivered kitDDPNotification was null for " + kit.getExternalOrderNumber());
+
+                if (!isReturn) {
```
zyme, line 194: Not much is really changed here--just passing in a `Connection` and adjusting accordingly.

src/main/java/org/broadinstitute/dsm/model/KitRequestExternal.java
```diff
@@ -29,28 +29,20 @@ public KitRequestExternal(String dsmKitRequestId, String participantId, String s
...
     // update kit request with status and date of external shipper
-    public static void updateKitRequest(String externalOrderStatus, long externalOrderDate, String dsmKitRequestId) {
-        SimpleResult results = inTransaction((conn) -> {
-            SimpleResult dbVals = new SimpleResult();
-            try (PreparedStatement stmt = conn.prepareStatement(SQL_UPDATE_KIT_REQUEST_EXTERNAL_SHIPPER_STATUS)) {
-                stmt.setString(1, externalOrderStatus);
-                stmt.setLong(2, externalOrderDate);
-                stmt.setString(3, dsmKitRequestId);
-                stmt.setString(4, externalOrderStatus);
+    public static void updateKitRequest(Connection conn, String externalOrderStatus, long externalOrderDate, String dsmKitRequestId) {
```
zyme, line 32: Another pass-in-the-Connection change is the only thing going on here.

src/main/java/org/broadinstitute/dsm/util/externalShipper/ExternalShipper.java
```diff
@@ -12,12 +12,14 @@
...
+     * Checks for incomplete orders and updates the internal status
+     * of the order
+     */
+    void updateOrderStatusForPendingKitRequests(int instanceId);
 
     public void orderConfirmation(long startDate, long endDate) throws Exception;
 
     public void orderCancellation(ArrayList<KitRequest> kitRequests) throws Exception;
 
-    public  ArrayList<KitRequest> getKitRequestsNotDone(int instanceId);
```
zyme, line 21: I moved this to `GBFRequestUtil`, rather than the interface, since it's now used only when testing `GBFRequestUtil`.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -199,65 +201,52 @@ else if (address != null) {
...
-                        KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(SQL_SELECT_SENT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER, kit.getExternalOrderNumber(), 2);//todo change this to the number of subkits but for now 2 for test boston works
-                        if (status.getOrderStatus().equals(NOT_FOUND) && StringUtils.isNotBlank(kit.getExternalOrderStatus()) && kit.getExternalOrderStatus().equals("NOT FOUND")
-                                && System.currentTimeMillis() - kit.getExternalOrderDate() >= TimeUnit.HOURS.toMillis(24)) {
-                            List<String> dsmKitRequestIds = getDSMKitRequestIds(status.getOrderNumber());
-                            if (dsmKitRequestIds != null && !dsmKitRequestIds.isEmpty()) {
-                                for (String dsmKitRequestId : dsmKitRequestIds) {
-                                    KitRequestExternal.updateKitRequest(status.getOrderStatus(), System.currentTimeMillis(), dsmKitRequestId);// in order to update time for the  next 24 hour check we need this
-                                }
+    public void orderStatus(Connection conn, KitRequest kit) throws Exception {
```
zyme, line 204: Rather than iterating on a list, this now operates on a single kit at a time, using the passed-in `Connection`.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -338,8 +327,59 @@ public void orderConfirmation(long startDate, long endDate) throws Exception {
...
+                        String ddpParticipantId = rs.getString(DBConstants.DDP_PARTICIPANT_ID);
+                        if (StringUtils.isNotBlank(ddpParticipantId)) {
+                            KitRequest kitRequest = new KitRequest(rs.getString(DBConstants.DSM_KIT_REQUEST_ID), ddpParticipantId,
+                                    null, null, rs.getString(DBConstants.EXTERNAL_ORDER_NUMBER), null,
+                                    rs.getString(DBConstants.EXTERNAL_ORDER_STATUS),
+                                    rs.getString("subkits." + DBConstants.EXTERNAL_KIT_NAME),
+                                    rs.getLong(DBConstants.EXTERNAL_ORDER_DATE));
+                            numOrdersProcessed.incrementAndGet();
+                            try {
+                                orderStatus(conn, kitRequest);
```
zyme, line 359: Wrapping this in a try/catch so we can keep going even if we run into a problem for a single `kitRequest`.

src/test/java/org/broadinstitute/dsm/QuartzExternalShipperTest.java
```diff
@@ -64,7 +64,7 @@ public static void last() {
     public void externalShipperJob() throws Exception {
 //        uploadKit();
 
-        ExternalShipper shipper = (ExternalShipper) Class.forName(DSMServer.getClassName("gbf")).newInstance();//GBFRequestUtil
+        GBFRequestUtil shipper = new GBFRequestUtil();
```
zyme, line 67: No need to instantiate via reflection, since we know exactly the `GBFRequestUtil` class we're testing here.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -199,65 +201,52 @@ else if (address != null) {
...
-                    orderNumbers.add(kit.getExternalOrderNumber());
-                    externalOrdersStatus.put(kit.getExternalOrderNumber(), kit);
-                }
-            }
-            JSONObject payload = new JSONObject().put("orderNumbers", orderNumbers);
-            logger.info("payload size for order status: " + orderNumbers.size());
-            String sendRequest = DSMServer.getBaseUrl(getExternalShipperName()) + STATUS_ENDPOINT;
-            logger.info("payload: " + payload.toString());
-            Response gbfResponse = executePost(Response.class, sendRequest, payload.toString(), DSMServer.getApiKey(getExternalShipperName()));
-            logger.info(gbfResponse.getXML());
```
zyme, line 217: This might be blowing things up--big log of a potentially very large string of XML.

pegahtah (approved):

PR #82, opened 2021-01-04
DDP-5479 Reducing the number of calls we make to GBF status endpoint

zyme:
For DDP-5479, we can reduce the number of calls we make to GBF status endpoint by:
* ignoring things that have been ordered in CE
* ignoring UPS status--it's not relevant here
* eliminating redundant calls for the same kit

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
@@ -57,8 +57,8 @@ public class GBFRequestUtil implements ExternalShipper {
             "                        req.kit_type_id, req.external_order_status, req.external_order_number, req.external_order_date, req.external_response, kt.no_return, req.created_by, kit.kit_label FROM kit_type kt, ddp_kit_request req, ddp_kit kit, ddp_instance ddp_site\n" +
             "                        WHERE req.ddp_instance_id = ddp_site.ddp_instance_id AND req.kit_type_id = kt.kit_type_id AND kit.dsm_kit_request_id = req.dsm_kit_request_id\n" +
             "                        and kit.test_result is null and kit.CE_order is null\n" +
-            "                        and not (kit.ups_tracking_status is not null\n" +
-            "                                and kit.ups_return_status is not null)\n" +
+            "                        and (kit.ups_tracking_status is null\n" +
+            "                             or kit.ups_return_status is null)\n" +
             "                        ) AS request\n" +
             "                        LEFT JOIN ddp_participant_exit ex ON (ex.ddp_instance_id = request.ddp_instance_id AND ex.ddp_participant_id = request.ddp_participant_id)\n" +
             "                        LEFT JOIN (SELECT subK.kit_type_id, subK.external_name from ddp_kit_request_settings dkc\n" +
--- a/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
@@ -63,7 +63,8 @@ public class GBFRequestUtil implements ExternalShipper {
             "                        LEFT JOIN ddp_participant_exit ex ON (ex.ddp_instance_id = request.ddp_instance_id AND ex.ddp_participant_id = request.ddp_participant_id)\n" +
             "                        LEFT JOIN (SELECT subK.kit_type_id, subK.external_name from ddp_kit_request_settings dkc\n" +
             "                        LEFT JOIN sub_kits_settings subK ON (subK.ddp_kit_request_settings_id = dkc.ddp_kit_request_settings_id)) as subkits ON (subkits.kit_type_id = request.kit_type_id)\n" +
-            "                    WHERE ex.ddp_participant_exit_id is null AND request.ddp_instance_id = ? AND NOT external_order_status <=> 'CANCELLED' AND external_order_number IS NOT NULL ";
+            "                    WHERE ex.ddp_participant_exit_id is null AND request.ddp_instance_id = ? AND (external_order_status not like '%CANCELLED%' or external_order_status is null) AND external_order_number IS NOT NULL\n" +
+            "                    order by request.dsm_kit_request_id desc ";
 
 
     private static final String SQL_SELECT_KIT_REQUEST_BY_EXTERNAL_ORDER_NUMBER = "SELECT dsm_kit_request_id FROM ddp_kit_request req WHERE external_order_number = ?";
@@ -355,6 +356,8 @@ public class GBFRequestUtil implements ExternalShipper {
                                     rs.getString("subkits." + DBConstants.EXTERNAL_KIT_NAME),
                                     rs.getLong(DBConstants.EXTERNAL_ORDER_DATE));
                             try {
+                                // keep track of which request ids we've already asked about, since this
+                                // result set may show subkits with the same external order id
                                 if (!queriedOrderIds.contains(kitRequest.getExternalOrderNumber())) {
                                     orderStatus(conn, kitRequest);
                                     numOrdersProcessed.incrementAndGet();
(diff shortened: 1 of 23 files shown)


src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -52,20 +52,19 @@
 
     private static final Logger logger = LoggerFactory.getLogger(GBFRequestUtil.class);
 
-    public static final String SQL_SELECT_EXTERNAL_KIT_NOT_DONE = "SELECT * FROM (SELECT kt.kit_type_name, ddp_site.instance_name, ddp_site.ddp_instance_id, ddp_site.base_url, ddp_site.auth0_token, ddp_site.migrated_ddp, " +
-            "            ddp_site.collaborator_id_prefix, req.bsp_collaborator_sample_id, req.ddp_participant_id, req.ddp_label, req.dsm_kit_request_id, req.bsp_collaborator_participant_id, " +
-            "            req.kit_type_id, req.external_order_status, req.external_order_number, req.external_order_date, req.external_response, kt.no_return, req.created_by, kit.kit_label FROM kit_type kt, ddp_kit_request req, ddp_kit kit, ddp_instance ddp_site " +
-            "            WHERE req.ddp_instance_id = ddp_site.ddp_instance_id AND req.kit_type_id = kt.kit_type_id AND kit.dsm_kit_request_id = req.dsm_kit_request_id " +
-            "            and kit.test_result is null " +
-            "            and not (kit.ups_tracking_status like 'D%' " +
```
zyme: Eliminate the requirement that kits be delivered.  With this, we ask GBF about things it has already told us have shipped until the kit is delivered, which isn't necessary.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -52,20 +52,19 @@
 
     private static final Logger logger = LoggerFactory.getLogger(GBFRequestUtil.class);
 
-    public static final String SQL_SELECT_EXTERNAL_KIT_NOT_DONE = "SELECT * FROM (SELECT kt.kit_type_name, ddp_site.instance_name, ddp_site.ddp_instance_id, ddp_site.base_url, ddp_site.auth0_token, ddp_site.migrated_ddp, " +
-            "            ddp_site.collaborator_id_prefix, req.bsp_collaborator_sample_id, req.ddp_participant_id, req.ddp_label, req.dsm_kit_request_id, req.bsp_collaborator_participant_id, " +
-            "            req.kit_type_id, req.external_order_status, req.external_order_number, req.external_order_date, req.external_response, kt.no_return, req.created_by, kit.kit_label FROM kit_type kt, ddp_kit_request req, ddp_kit kit, ddp_instance ddp_site " +
-            "            WHERE req.ddp_instance_id = ddp_site.ddp_instance_id AND req.kit_type_id = kt.kit_type_id AND kit.dsm_kit_request_id = req.dsm_kit_request_id " +
-            "            and kit.test_result is null " +
```
zyme: We can also ignore things that have been ordered in CE, not just those that have had results.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -52,20 +52,20 @@
...
+            "                        and kit.test_result is null and kit.CE_order is null\n" +
+            "                        and not (kit.ups_tracking_status is not null\n" +
+            "                                and kit.ups_return_status is not null)\n" +
+            "                        ) AS request\n" +
+            "                        LEFT JOIN ddp_participant_exit ex ON (ex.ddp_instance_id = request.ddp_instance_id AND ex.ddp_participant_id = request.ddp_participant_id)\n" +
+            "                        LEFT JOIN (SELECT subK.kit_type_id, subK.external_name from ddp_kit_request_settings dkc\n" +
+            "                        LEFT JOIN sub_kits_settings subK ON (subK.ddp_kit_request_settings_id = dkc.ddp_kit_request_settings_id)) as subkits ON (subkits.kit_type_id = request.kit_type_id)\n" +
+            "                    WHERE ex.ddp_participant_exit_id is null AND request.ddp_instance_id = ? AND (external_order_status not like '%CANCELLED%' or external_order_status is null) AND external_order_number IS NOT NULL\n" +
+            "                    order by request.dsm_kit_request_id desc ";
```
zyme: Adding an order so we are always looking at most recent first, since the fresher a row is, the more likely that it's status is changing.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -52,20 +52,20 @@
...
-            "            and kit.test_result is null " +
-            "            and not (kit.ups_tracking_status like 'D%' " +
-            "                    and kit.ups_return_status like 'D%' " +
-            "                    and kit.ups_tracking_status is not null " +
-            "                    and kit.ups_return_status is not null) " +
-            "            ) AS request " +
-            "            LEFT JOIN ddp_participant_exit ex ON (ex.ddp_instance_id = request.ddp_instance_id AND ex.ddp_participant_id = request.ddp_participant_id) " +
-            "            LEFT JOIN (SELECT subK.kit_type_id, subK.external_name from ddp_kit_request_settings dkc " +
-            "            LEFT JOIN sub_kits_settings subK ON (subK.ddp_kit_request_settings_id = dkc.ddp_kit_request_settings_id)) as subkits ON (subkits.kit_type_id = request.kit_type_id) " +
-            "            WHERE ex.ddp_participant_exit_id is null AND request.ddp_instance_id = ? AND NOT external_order_status <=> 'CANCELLED' AND external_order_number IS NOT NULL " ;
```
zyme: Changed the check for `CANCELLED` to capture more variations on `CANCELLED`, not just a strict equals, since we have more than one `CANCELLED`.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -52,20 +52,20 @@
...
+    public static final String SQL_SELECT_EXTERNAL_KIT_NOT_DONE = "SELECT * FROM (SELECT kt.kit_type_name, ddp_site.instance_name, ddp_site.ddp_instance_id, ddp_site.base_url, ddp_site.auth0_token, ddp_site.migrated_ddp,\n" +
+            "                        ddp_site.collaborator_id_prefix, req.bsp_collaborator_sample_id, req.ddp_participant_id, req.ddp_label, req.dsm_kit_request_id, req.bsp_collaborator_participant_id,\n" +
+            "                        req.kit_type_id, req.external_order_status, req.external_order_number, req.external_order_date, req.external_response, kt.no_return, req.created_by, kit.kit_label FROM kit_type kt, ddp_kit_request req, ddp_kit kit, ddp_instance ddp_site\n" +
+            "                        WHERE req.ddp_instance_id = ddp_site.ddp_instance_id AND req.kit_type_id = kt.kit_type_id AND kit.dsm_kit_request_id = req.dsm_kit_request_id\n" +
+            "                        and kit.test_result is null and kit.CE_order is null\n" +
+            "                        and not (kit.ups_tracking_status is not null\n" +
```
SimoneMaiwald: should we maybe change that to 
"kit.tracking_to_id is null or kit.tracking_return_id is null or kit.kit_label is null"
and loose the "not not" situation. because if we have all of these 3 ids, it means we already got the kit per confirmation and that the status must be shipped?

zyme: Refining this a bit more--we don't need ups status at all.  We can just check kit label and tracking ids.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -51,19 +53,18 @@
...
-            "            WHERE ex.ddp_participant_exit_id is null AND request.ddp_instance_id = ? AND NOT external_order_status <=> 'CANCELLED' AND external_order_number IS NOT NULL\n";
+            "                                    ddp_site.collaborator_id_prefix, req.bsp_collaborator_sample_id, req.ddp_participant_id, req.ddp_label, req.dsm_kit_request_id, req.bsp_collaborator_participant_id,\n" +
+            "                                    req.kit_type_id, req.external_order_status, req.external_order_number, req.external_order_date, req.external_response, kt.no_return, req.created_by, kit.kit_label FROM kit_type kt, ddp_kit_request req, ddp_kit kit, ddp_instance ddp_site\n" +
+            "                                    WHERE req.ddp_instance_id = ddp_site.ddp_instance_id AND req.kit_type_id = kt.kit_type_id AND kit.dsm_kit_request_id = req.dsm_kit_request_id\n" +
+            "                                    and kit.test_result is null and kit.CE_order is null\n" +
+            "                                    and (kit.kit_label is null or kit.tracking_return_id is null or kit.tracking_to_id is null))\n" +
```
zyme, line 60: @SimoneMaiwald highlighting this change so we don't lose track of it in this PR.

SimoneMaiwald (approved):

PR #84, opened 2021-01-05
DDP-5479: java heap dump

SimoneMaiwald:
ParticipantWrapper test change to see if it would help to filter ES participants if list was filtered for a kit

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/DSMServer.java
+++ b/src/main/java/org/broadinstitute/dsm/DSMServer.java
@@ -334,9 +334,10 @@ public class DSMServer extends BasicServer {
         get(UI_ROOT + "/heapDump", new Route() {
             @Override
             public Object handle(Request request, Response response) throws Exception {
-            String gcpName = TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.GOOGLE_PROJECT_NAME);
-            heapDumper.dumpHeapToBucket(gcpName + "_dsm_heapdumps");
-            return null;
+                logger.info("Received request to create java heap dump");
+                String gcpName = TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.GOOGLE_PROJECT_NAME);
+                heapDumper.dumpHeapToBucket(gcpName + "_dsm_heapdumps");
+                return null;
             }
         }, new JsonTransformer());
         logger.info("Finished setting up DSM custom routes and jobs...");
--- a/src/main/java/org/broadinstitute/dsm/DSMServer.java
+++ b/src/main/java/org/broadinstitute/dsm/DSMServer.java
@@ -329,6 +329,16 @@ public class DSMServer extends BasicServer {
 
         setupJobs(cfg, kitUtil, notificationUtil, eventUtil, container, receiver);
 
+        //TODO - redo with pubsub
+        JavaHeapDumper heapDumper = new JavaHeapDumper();
+        get(UI_ROOT + "/heapDump", new Route() {
+            @Override
+            public Object handle(Request request, Response response) throws Exception {
+            String gcpName = TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.GOOGLE_PROJECT_NAME);
+            heapDumper.dumpHeapToBucket(gcpName + "_heapdumps");
+            return null;
+            }
+        }, new JsonTransformer());
         logger.info("Finished setting up DSM custom routes and jobs...");
     }
 
--- a/src/main/java/org/broadinstitute/dsm/DSMServer.java
+++ b/src/main/java/org/broadinstitute/dsm/DSMServer.java
@@ -335,7 +335,7 @@ public class DSMServer extends BasicServer {
             @Override
             public Object handle(Request request, Response response) throws Exception {
             String gcpName = TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.GOOGLE_PROJECT_NAME);
-            heapDumper.dumpHeapToBucket(gcpName + "_heapdumps");
+            heapDumper.dumpHeapToBucket(gcpName + "_dsm_heapdumps");
             return null;
             }
         }, new JsonTransformer());
--- a/src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
+++ b/src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
@@ -25,6 +25,9 @@ public class ParticipantWrapper {
 
     private static final Logger logger = LoggerFactory.getLogger(ParticipantWrapper.class);
 
+    public static final String BY_DDP_PARTICIPANT_ID_IN = " AND request.ddp_participant_id IN (\"";
+    public static final String ORDER_AND_LIMIT = " ORDER BY request.dsm_kit_request_id desc LIMIT 5000";
+
     private Map<String, Object> data;
     private Participant participant;
     private List<MedicalRecord> medicalRecords;
@@ -63,7 +66,7 @@ public class ParticipantWrapper {
                 oncHistoryDetails = OncHistoryDetail.getOncHistoryDetails(instance.getName());
             }
             if (DDPInstance.getRole(instance.getName(), DBConstants.KIT_REQUEST_ACTIVATED)) { //only needed if study is shipping samples per DSM
-                kitRequests = KitRequestShipping.getKitRequests(instance);
+                kitRequests = KitRequestShipping.getKitRequests(instance, ORDER_AND_LIMIT);
             }
             Map<String, List<AbstractionActivity>> abstractionActivities = AbstractionActivity.getAllAbstractionActivityByRealm(instance.getName());
             Map<String, List<AbstractionGroup>> abstractionSummary = AbstractionFinal.getAbstractionFinal(instance.getName());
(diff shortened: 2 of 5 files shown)


src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -329,6 +329,16 @@ protected void setupCustomRouting(@NonNull Config cfg) {
 
         setupJobs(cfg, kitUtil, notificationUtil, eventUtil, container, receiver);
 
+        //TODO - redo with pubsub
+        JavaHeapDumper heapDumper = new JavaHeapDumper();
+        get(UI_ROOT + "/heapDump", new Route() {
+            @Override
+            public Object handle(Request request, Response response) throws Exception {
+            String gcpName = TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.GOOGLE_PROJECT_NAME);
```
zyme: Maybe add a log statement to show that we're dumping the heap?

src/main/java/org/broadinstitute/dsm/util/JavaHeapDumper.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public void dumpHeapToBucket(String bucketName, final String fileName) throws IOException {
+        String gcpName = TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.GOOGLE_PROJECT_NAME);
+        dumpHeapToLocalFile(DEFAULT_LOCAL_PATH + "/" + fileName);
+        File localDumpFile = Paths.get(DEFAULT_LOCAL_PATH, fileName).toFile();
+        if (localDumpFile.exists()) {
+            logger.info("Created local dump file: " + localDumpFile.getAbsolutePath() + " with size: " + localDumpFile.length());
+        } else {
+            logger.error("Could not find dump file at:" + DEFAULT_LOCAL_PATH + "/" + fileName);
```
zyme: Probably don't need the logger.error since the next line throws an exception with similar info.

src/main/java/org/broadinstitute/dsm/DSMServer.java
zyme: Looks like existing auth filter based on `UI_ROOT` protects this, so yay!

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -124,7 +125,15 @@ else if (StringUtils.isNotBlank(instance.getParticipantIndexES())){
             }
             //get all the list which were not filtered
             if (participantESData == null) {
-                participantESData = getESData(instance);
+                //get only pts for the filtered kitRequests
+                if (kitRequests != null && !kitRequests.isEmpty()) {
```
zyme, line 132: This looks like it will help.  `KitRequestShipping.getKitRequests()` on L109 and L66 is still an issue, however.  It looks like that method will still pull in all kit requests for the study.  Is it possible to address this as well?  The underlying query is returning about 45k rows:

```
SELECT kt.kit_type_name, realm.instance_name, request.bsp_collaborator_participant_id, request.bsp_collaborator_sample_id, request.ddp_participant_id, request.ddp_label, request.dsm_kit_request_id, 
            request.kit_type_id, request.external_order_status, request.external_order_number, request.external_order_date, request.external_response, request.upload_reason, kt.no_return, request.created_by, 
            kit.dsm_kit_request_id, kit.dsm_kit_id, kit.kit_complete, kit.label_url_to, kit.label_url_return, kit.tracking_to_id, 
            kit.tracking_return_id, kit.easypost_tracking_to_url, kit.easypost_tracking_return_url, kit.easypost_to_id, kit.easypost_shipment_status, kit.scan_date, kit.label_date, kit.error, kit.message, 
            kit.receive_date, kit.deactivated_date, kit.easypost_address_id_to, kit.deactivation_reason, (select t.tracking_id from ddp_kit_tracking t where t.kit_label = kit.kit_label) as tracking_id, kit.kit_label, kit.express, kit.test_result, kit.needs_approval, kit.authorization, kit.denial_reason, 
            kit.authorized_by, kit.ups_tracking_status, kit.ups_return_status, kit.CE_order 
            FROM ddp_kit_request request 
            LEFT JOIN ddp_kit kit on (kit.dsm_kit_request_id = request.dsm_kit_request_id) 
            LEFT JOIN ddp_instance realm on (realm.ddp_instance_id = request.ddp_instance_id) 
            LEFT JOIN kit_type kt on (request.kit_type_id = kt.kit_type_id)  where realm.instance_name = 'testboston'
 ```

File: src/main/java/org/broadinstitute/dsm/util/JavaHeapDumper.java [hunk 4]
```diff
@@ -0,0 +1,84 @@
...
+        String serviceName = System.getenv("GAE_SERVICE");
+        String instanceId = System.getenv("GAE_INSTANCE");
+        String dateTime = DateTimeFormatter.ISO_DATE_TIME.format(LocalDateTime.now());
+        return appName != null ? appName + "-" + serviceName + "-" + instanceId + "-" + dateTime + ".hprof" : null;
+    }
+
+    private String nonGcpFileName() {
+        return "heapDump-" + DateTimeFormatter.ISO_DATE_TIME.format(LocalDateTime.now()) + ".hprof";
+    }
+}
```
MocanaAtBroad, line 83: Heap dump code looks good.

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
zyme: I think we may have to do something like set `limit 1000` and order by something like `user_id desc` in `KitRequestShipping.getKitRequests()` when there are no filters passed in.  Otherwise, we are going to read in everything into memory.  It's a drastic limitation, but short of rearchitecting things so that **all** data that feeds these views goes into ES and queries **only** stream out of ES (avoiding app server memory entirely with either UI redirects or some such), I don't see an elegant or quick way to ensure that DSM won't blow out memory limits.

zyme: Come to think of it, maybe we should apply this kind of limit across the board until we come up with something better--no matter what, we cap the size of results at 1k?  5k?

MocanaAtBroad (commented):
Would highly recommend to treat the ElasticSearch client as a singleton instead of creating new ones on each request. Each client has a thread pool and each thread that execute will consume about 1MB of memory.

zyme (approved):
A few minor suggestions--let's roll this out and address `KitRequestShipping` separately.

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
SimoneMaiwald: I will add the filtering to L109 and L66 if the list was filtered for something and if not I will add a default cap of results and a comment to the UI that if something is missing user should filter for that pt to get "all" informations

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -63,7 +66,7 @@ public JsonObject getDataAsJson() {
                 oncHistoryDetails = OncHistoryDetail.getOncHistoryDetails(instance.getName());
             }
             if (DDPInstance.getRole(instance.getName(), DBConstants.KIT_REQUEST_ACTIVATED)) { //only needed if study is shipping samples per DSM
-                kitRequests = KitRequestShipping.getKitRequests(instance);
+                kitRequests = KitRequestShipping.getKitRequests(instance, ORDER_AND_LIMIT);
```
zyme, line 69: Nice, and it looks like `KitRequestShipping.getKitRequests(@NonNull DDPInstance instance)` is unused so we could remove it!

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -145,7 +148,15 @@ else if (StringUtils.isNotBlank(instance.getParticipantIndexES())){
                 oncHistories = OncHistoryDetail.getOncHistoryDetails(instance.getName());
             }
             if (kitRequests == null && DDPInstance.getRole(instance.getName(), DBConstants.KIT_REQUEST_ACTIVATED)) { //only needed if study is shipping samples per DSM
-                kitRequests = KitRequestShipping.getKitRequests(instance);
+                //get only kitRequests for the filtered pts
+                if (participantESData != null && !participantESData.isEmpty()) {
+                    String filter = Arrays.stream(participantESData.keySet().toArray(new String[0])).collect(Collectors.joining("\",\""));
```
zyme, line 153: So we'll be making string here that is basically of size (lenghOfGuid * numberOfParticipants)?  That might be an issue as we approach 15k participants, but let's give it a try.

Also, can you check out the limits for [mysql IN clause](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_in)? Looks like:

> The number of values in the IN() list is only limited by the max_allowed_packet value

Currently in prod this is 33554432.

zyme: Maybe throw in a log statement so we can keep an eye on sizes?

```
logger.info("About to query for " + kitRequests.size() + " kits");
```

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
SimoneMaiwald: I will keep it for now in case we need it later…

PR #85, opened 2021-01-05
adding a check to not update data of other kit

SimoneMaiwald:


src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -38,9 +38,9 @@
 
 
     private static final String SQL_UPDATE_UPS_TRACKING_STATUS = "UPDATE ddp_kit SET ups_tracking_status = ?, ups_tracking_date = ? " +
-            "WHERE dsm_kit_id <> 0 and  tracking_to_id = ? ";
+            "WHERE dsm_kit_id <> 0 and  tracking_to_id = ? and test_result is null";
```
SimoneMaiwald, line 40: added assumption that if a kit already has test_results it can not be the one we are currently checking for tracking information

zyme: Could also include `or ce_ordered is null`, since that will capture things that have returned, since sometimes there are no results...

PR #86, opened 2021-01-06
DDP-5479: refactoring ups job

SimoneMaiwald:
- streaming instead of big list
- update sql will check for external order number and not just tracking number

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
@@ -50,8 +50,7 @@ public class TestBostonUPSTrackingJob implements Job {
     static String DELIVERED = "DELIVERED";
     static String RECEIVED = "RECEIVED";
 
-    private static String SELECT_BY_TRACKING_NUMBER = "and kit.tracking_to_id = ?";
-    private static String SELECT_BY_RETURN_NUMBER = "and kit.tracking_return_id = ?";
+    private static String SELECT_BY_EXTERNAL_ORDER_NUMBER = "and req.external_order_number = ?";
     private static Covid19OrderRegistrar orderRegistrar;
 
     @Override
@@ -214,7 +213,7 @@ public class TestBostonUPSTrackingJob implements Job {
 
             if (!isReturn) {
                 if (shouldTriggerEventForKitOnItsWayToParticipant(statusType, oldType)) {
-                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(conn,SQL_SELECT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER + SELECT_BY_TRACKING_NUMBER, new String[] { DELIVERED, trackingId }, 2);//todo change this to the number of subkits but for now 2 for test boston works
+                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(conn,SQL_SELECT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER + SELECT_BY_EXTERNAL_ORDER_NUMBER, new String[] { DELIVERED, kit.getExternalOrderNumber() }, 2);//todo change this to the number of subkits but for now 2 for test boston works
                     if (kitDDPNotification != null) {
                         logger.info("Triggering DDP for kit going to participant with external order number: " + kit.getExternalOrderNumber());
                         EventUtil.triggerDDP(conn, kitDDPNotification);
@@ -239,7 +238,7 @@ public class TestBostonUPSTrackingJob implements Job {
                 if (shouldTriggerEventForReturnKitDelivery(statusType, oldType)) {
                     KitUtil.setKitReceived(conn, kit.getMainKitLabel());
                     logger.info("RECEIVED: " + trackingId);
-                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(conn,SQL_SELECT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER + SELECT_BY_RETURN_NUMBER, new String[] { RECEIVED, trackingId }, 2);//todo change this to the number of subkits but for now 2 for test boston works
+                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(conn,SQL_SELECT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER + SELECT_BY_EXTERNAL_ORDER_NUMBER, new String[] { RECEIVED, kit.getExternalOrderNumber() }, 2);//todo change this to the number of subkits but for now 2 for test boston works
                     if (kitDDPNotification != null) {
                         logger.info("Triggering DDP for received kit with external order number: " + kit.getExternalOrderNumber());
                         EventUtil.triggerDDP(conn, kitDDPNotification);
--- a/src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsm.jobs;
 
+import lombok.NonNull;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.ddp.db.SimpleResult;
 import org.broadinstitute.dsm.DSMServer;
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -36,10 +37,10 @@
     private static String SQL_AVOID_DELIVERD = " and (ups_tracking_status is null or ups_return_status is null or ups_tracking_status not like \"" + UPSStatus.DELIVERED_TYPE + " %\" or ups_return_status not like \"" + UPSStatus.DELIVERED_TYPE + " %\")" +
             " order by kit.dsm_kit_request_id ASC";
 
-    private static final String SQL_UPDATE_UPS_TRACKING_STATUS = "UPDATE ddp_kit SET ups_tracking_status = ?, ups_tracking_date = ? " +
-            "WHERE dsm_kit_id <> 0 and  tracking_to_id = ? and test_result is null";
-        private static final String SQL_UPDATE_UPS_RETURN_STATUS = "UPDATE ddp_kit SET ups_return_status = ?, ups_return_date = ? " +
-                "WHERE dsm_kit_id <> 0 and tracking_return_id= ? and test_result is null";
+    private static final String SQL_UPDATE_UPS_TRACKING_STATUS = "UPDATE ddp_kit kit, ddp_kit_request request SET ups_tracking_status = ?, ups_tracking_date = ? " +
```
zyme: I think this should take the form `UPDATE ddp_kit where...` and avoid putting `ddp_kit_request` into the update statement.  I'm not sure mysql is smart enough to **not** put an update lock on `ddp_kit_request`.  It might be, but better to make the update as narrow as possible.

Also, are `tracking_to_id` and `tracking_return_id` indexed?  If not, please index them so that these update locks are held as briefly as possible.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -58,36 +59,71 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
...
+                                        rs.getString("req." + DBConstants.EXTERNAL_ORDER_NUMBER),
+                                        rs.getBoolean("kit." + DBConstants.CE_ORDER)
+                                );
+                                String type;
+                                if (StringUtils.isNotBlank(kit.getTrackingToId())) {
+                                    type = kit.getUpsTrackingStatus();
+                                    if (StringUtils.isNotBlank(type)) {// get only type from it
+                                        type = type.substring(0, type.indexOf(' '));
+                                    }
+                                    if (!"D".equals(type)) {//don't include delivered ones
```
zyme: Please use `UPSStatus.DELIVERED_TYPE` instead of `"D"` so that we can find dependencies more easily.  Better still might be a static method on `UPSStatus` like `isDelivered(String statusType)`.

zyme: Actually even better would be to also add `isDelivered()` to `DdpKit`.  It can use `UPSStatus.isDelivered` internally, and then the information about what you need is right there.  Check out Feature Envy [here](https://blog.codinghorror.com/code-smells/) and [here](https://softwareengineering.stackexchange.com/questions/212128/what-is-a-feature-envy-code-and-why-is-it-considered-a-code-smell).

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -58,36 +59,71 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
...
-                        }
-                        kits = ids.get("return");
-                        if (kits != null) {
-                            logger.info("checking return status for " + kits.size() + " tracking numbers");
-                            for (DdpKit kit : kits.values()) {
-                                try {
-                                    updateKitStatus(kit, true);
-                                } catch (Exception e) {
-                                    logger.error("Could not update return status for " + kit.getExternalOrderNumber(), e);
+                                if (StringUtils.isNotBlank(kit.getTrackingReturnId())) {
```
zyme: Similar comment here--by putting the defensive parse/null checks into `DdpKit.isDelivered()` (or expanding to two methods like `hasArrived()` and `hasReturned()`, you can centralize this, test it, and not have to duplicate it in other places.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -58,36 +59,71 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
...
+                                    type = kit.getUpsReturnStatus();
+                                    if (StringUtils.isNotBlank(type)) {
+                                        type = type.substring(0, type.indexOf(' '));
+                                    }
+                                    if (!"D".equals(type)) {
+                                        if (kit.getKitLabel().contains("_1") && kit.getKitLabel().indexOf("_1") == kit.getKitLabel().length() - 2) {
```
zyme: Magic string alert...what is `"_1"`, and what does it mean business-wise?  Instead of dipping into the label and doing an a check on it, write a method that tells you what you need to know.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -180,61 +216,59 @@ private static void updateTrackingInfo(String statusType,
...
-                        }
+            if (!isReturn) {
+                if (shouldTriggerEventForKitOnItsWayToParticipant(statusType, oldType)) {
+                    KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(conn,SQL_SELECT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER + SELECT_BY_TRACKING_NUMBER, new String[] { DELIVERED, trackingId }, 2);//todo change this to the number of subkits but for now 2 for test boston works
+                    if (kitDDPNotification != null) {
+                        logger.info("Triggering DDP for kit going to participant with external order number: " + kit.getExternalOrderNumber());
+                        EventUtil.triggerDDP(conn, kitDDPNotification);
                     }
+                    else {
+                        logger.error("delivered kitDDPNotification was null for " + kit.getExternalOrderNumber());
```
zyme, line 222: What does nullness here imply at a higher level?  If it implies that a certain thing is broken and needs tending to, please bake that into the comment.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
SimoneMaiwald: @pegahtah can you double check why you checked if label has `_1` 
if it is `_1` you remove it and update the label to substring without `_1`

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
SimoneMaiwald: @pegahtah can you look into that please

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -180,61 +202,59 @@ private static void updateTrackingInfo(String statusType,
...
                     }
+                }
 
+            }
+            else { // this is a return
+                if (earliestInTransitTime != null && !kit.isCEOrdered()) {
+                    // if we have the first date of an inbound event, create an order in CE
+                    // using the earliest date of inbound event
+                    orderRegistrar.orderTest(DSMServer.careEvolveAuth, kit.getHRUID(), kit.getMainKitLabel(), kit.getExternalOrderNumber(), earliestInTransitTime);
```
SimoneMaiwald, line 231: @pegahtah is this and L240 the only places where it is important that kitLabel is not `_1`

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
pegahtah: @SimoneMaiwald a null ness here means a difference in data between ES and DSM. It some times happen in test environments where test users gets deleted but I don't think we will really face it in prod.

However, based on what we learned maybe we should change the SELECT query to not use `SELECT_BY_TRACKING_NUMBER` since now we can have multiple?

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
pegahtah: Yes rest share the same info only where kit label is  needed for a method we need it to be the main one

pegahtah (approved):

PR #90, opened 2021-01-08
Merge remote-tracking branch 'origin/appengine' into prod_01072021

pegahtah:
We need to filter our kit selection based on the test_result being null because otherwise we get kits that are already delivered to us but ups didn't update their status


SimoneMaiwald:
@pegahtah can you check that it is in develop and then close the pr

PR #92, opened 2021-01-11
GBFStatus endpoint bad format fix

pegahtah:


src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -202,7 +202,9 @@ else if (address != null) {
     // return the actual tube barcode for all tubes but only after the kit is shipped (around 7:00pm)
     public void orderStatus(Connection conn, KitRequest kit) throws Exception {
         logger.info("checking status of " + kit.getExternalOrderNumber() + " for participant " + kit.getParticipantId());
-        JSONObject payload = new JSONObject().put("orderNumbers", kit.getExternalOrderNumber());
+        List<String> orderNumbers = new ArrayList<>();
+        orderNumbers.addAll(Arrays.asList(new String[] { kit.getExternalOrderNumber() }));
+        JSONObject payload = new JSONObject().put("orderNumbers", orderNumbers);
```
pegahtah, line 207: We were getting bad formats in terms of `gbfResponse.isSuccess()` being false

PR #93, opened 2021-01-13
Cloud Function POC

zyme:
Something we should consider is offloading quartz jobs to cloud functions so that our API resources don't get dragged down by background jobs.  This is a POC that shows how a cloud function can be built and deployed with (hopefully) minimal impact on the current process for deploying the backend.  Early drafts used separate `pom.xml` files, but with creative use of [maven profiles](https://maven.apache.org/guides/introduction/introduction-to-profiles.html), we can use a single `pom.xml`.

[TestBostonKitTrackerDispatcher](https://github.com/broadinstitute/ddp-study-manager/pull/93/files#diff-8abf39e1f4eabae2c5323b9335964051e50098ae0831020915f9e4cd2244e548) does one thing: grabs a list of to/from UPS tracking numbers and prints them.

Because [cloud functions may cache and re-use `static` vars](https://cloud.google.com/functions/docs/bestpractices/tips#functions-tips-scopes-java) across invocations, I had to tease out a separate connection methodology.  There are probably more things like this that we will encounter as we move more things into cloud functions.

Cloud functions also have [hard limits](https://cloud.google.com/functions/quotas) in terms of memory and compute time, so `TestBostonKitTrackerDispatcher` is probably just one piece that creates a worklist, since generating 20,000 calls to UPS will probably take more than nine minutes.  Architecting our cloud functions to fit within these resource constraints will be a new and exciting endeavor!

Check out the [CF in GCP](https://console.cloud.google.com/functions/list?project=broad-ddp-dev&tab=logs).  You can kick the tires and view the logs.

@Irakli-Khaladze curious to get your feedback on this in the context of DDP-5445.

To build, be sure to use `-Pcould-function` in mvn, both command-line and Intellij.  Note that this change will also require use of `-Papis`for building backend APIs.

### DIFF
--- a/build-and-deploy-cloud-functions.sh
+++ b/build-and-deploy-cloud-functions.sh
@@ -5,7 +5,7 @@ STUDY_SERVER_SCHEMA=$4
 
 echo "Will deploy to ${PROJECT_ID} using schemas ${STUDY_MANAGER_SCHEMA} and ${STUDY_SERVER_SCHEMA}"
 
-mvn -Pcloud-function -DskipTests clean install package
+# mvn -Pcloud-function -DskipTests clean install package
 
 
 #echo "Deploying kit dispatcher to ${PROJECT_ID}"
@@ -15,7 +15,9 @@ mvn -Pcloud-function -DskipTests clean install package
 #    --runtime=java11 \
 #    --trigger-topic=kit-report \
 #    --source=target/deployment \
-#    --set-env-vars="PROJECT_ID=${PROJECT_ID},SECRET_ID=cloud-functions,STUDY_MANAGER_SCHEMA=${STUDY_MANAGER_SCHEMA},STUDY_SERVER_SCHEMA=${STUDY_SERVER_SCHEMA}"
+#    --set-env-vars="PROJECT_ID=${PROJECT_ID},SECRET_ID=cloud-functions,STUDY_MANAGER_SCHEMA=${STUDY_MANAGER_SCHEMA},STUDY_SERVER_SCHEMA=${STUDY_SERVER_SCHEMA}" \
+#    --egress-settings=all \
+#    --vpc-connector=appengine-default-connect
 
 
 
@@ -27,4 +29,5 @@ gcloud --project=${PROJECT_ID} functions deploy \
     --trigger-topic=tbos-ce-orders \
     --source=target/deployment \
     --memory=1024MB \
-    --set-env-vars="PROJECT_ID=${PROJECT_ID},SECRET_ID=study-manager-config"
+    --set-env-vars="PROJECT_ID=${PROJECT_ID},SECRET_ID=study-manager-config" \
+    --vpc-connector=projects/broad-ddp-dev/locations/us-central1/connectors/appengine-default-connect
--- a/build-and-deploy-cloud-functions.sh
+++ b/build-and-deploy-cloud-functions.sh
@@ -1,0 +1,18 @@
+
+PROJECT_ID=$1
+SECRET_ID=$2
+STUDY_MANAGER_SCHEMA=$3
+STUDY_SERVER_SCHEMA=$4
+
+echo "Will deploy to ${PROJECT_ID} with ${SECRET_ID} secret using schemas ${STUDY_MANAGER_SCHEMA} and ${STUDY_SERVER_SCHEMA}"
+
+mvn -Pcloud-function -DskipTests clean install package
+
+echo "Deploying to ${PROJECT_ID}"
+gcloud --project=${PROJECT_ID} functions deploy \
+    tbos-kit-tracking-dispatcher \
+    --entry-point=org.broadinstitute.dsm.cf.TestBostonKitTrackerDispatcher \
+    --runtime=java11 \
+    --trigger-topic=kit-report \
+    --source=target/deployment \
+    --set-env-vars="PROJECT_ID=${PROJECT_ID},SECRET_ID=${SECRET_ID},STUDY_MANAGER_SCHEMA=${STUDY_MANAGER_SCHEMA},STUDY_SERVER_SCHEMA=${STUDY_SERVER_SCHEMA}"
--- a/build-and-deploy-cloud-functions.sh
+++ b/build-and-deploy-cloud-functions.sh
@@ -1,18 +1,30 @@
 
 PROJECT_ID=$1
-SECRET_ID=$2
 STUDY_MANAGER_SCHEMA=$3
 STUDY_SERVER_SCHEMA=$4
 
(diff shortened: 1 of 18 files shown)


src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,134 @@
...
+                "k.tracking_to_id,\n" +
+                "k.tracking_return_id,\n" +
+                "json_extract(k.test_result, '$[0].result') test_result,\n" +
+                "ifnull(STR_TO_DATE(replace(json_extract(k.test_result, '$[0].timeCompleted'),'\"',''), '%Y-%m-%dT%H:%i:%sZ'),\n" +
+                "    STR_TO_DATE(replace(json_extract(k.test_result, '$[0].timeCompleted'),'\"',''), '%Y-%m-%dT%H:%i:%s.%fZ')\n" +
+                ")as test_completion_time,\n" +
+                "u.hruid,\n" +
+                "dkr.upload_reason\n" +
+                "from\n" +
+                STUDY_MANAGER_SCHEMA + "ddp_kit k,\n" +
```
zyme, line 57: Making `KIT_QUERY` a static var may run into caching problems.  Having different schema names between dev and prod is great for minimizing the chance of accidents, but calling out the schemas differently here is a bit clunky.

pom.xml
```diff
@@ -234,54 +228,147 @@
...
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
     <build>
         <sourceDirectory>src/main/java</sourceDirectory>
         <testSourceDirectory>src/test/java</testSourceDirectory>
-        <resources>
```
zyme, line 249: Basically just moved this under the `apis` profile.

pom.xml
```diff
@@ -234,54 +228,147 @@
...
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-secretmanager</artifactId>
+        </dependency>
+
     </dependencies>
 
+    <profiles>
+        <profile>
+            <id>cloud-function</id>
+            <build>
```
zyme, line 254: This is lifted largely from GCPs tutorial on java cloud functions.

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,134 @@
...
+        try (Connection conn = dataSource.getConnection()) {
+            try (PreparedStatement stmt = conn.prepareStatement(KIT_QUERY)) {
+                stmt.setFetchSize(1000);
+                try (ResultSet rs = stmt.executeQuery()) {
+                    while (rs.next()) {
+                        String userGuid = rs.getString("guid");
+                        logger.info(userGuid);
+
+                        // spawn threads or other cloud functions in fixed size blocks? Write to a separate queue for worker threads
+                        // to grab history?
```
zyme, line 93: With CF limits, churning through the whole result set here will result in termination before the list is complete.  I'm leaning towards capturing these guids as elements of a generic worklist in json as a sort of IPC for this and other producer/consumer patterns.

A TBD class like `TestBostonKitTrackingConsumer` or some such could then run on a GCP schedule (let's say every 30m) and run through the queue, keeping track of its own execution time, shutting down gracefully before it is terminated.  In this way we would have sustained progress through the list without running afoul of CF limits, and without overwhelming UPS with many concurrent calls.

zyme:
Also worth mentioning--I think there is a nice opportunity here to deploy a bunch of one-off "scripts" or "tests" as CFs, and then trigger them via the CF "test" console, where we can lob in json objects.  I'm thinking of things like adhoc orders of covid tests, kits, etc...

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,134 @@
...
+ * Identifies kits whose shipping history should be
+ * queried (by another job) in order to produce
+ * reports on turnaround time and study adherence.
+ */
+public class TestBostonKitTrackerDispatcher implements BackgroundFunction<PubsubMessage> {
+
+    private static final Logger logger = Logger.getLogger(TestBostonKitTrackerDispatcher.class.getName());
+
+    private String STUDY_MANAGER_SCHEMA = System.getenv("STUDY_MANAGER_SCHEMA") + ".";
+    private String STUDY_SERVER_SCHEMA = System.getenv("STUDY_SERVER_SCHEMA") + ".";
```
yufengwng, line 37: Can we put the schema names into the config file? We're already reading the config file from SecretManager on execution so might as well put it in there, and then construct the kit query after pulling in config. Config file will be different for each environment, so we can have different values in there.

yufengwng (commented):
Didn't read the kit query carefully yet, but overall setup looks good to me.

Irakli-Khaladze:
I have one question, when we deploy cloud functions with these approach, are we going to have whole source code in the cloud as source of these functions?

zyme:
@Irakli-Khaladze yes, we'll have all the study-manager code in the CF.  Any concerns with this approach?

Irakli-Khaladze:
> @Irakli-Khaladze yes, we'll have all the study-manager code in the CF. Any concerns with this approach?

At first, I had impression that function would be separate from dsm and would have it's own source on cloud, but in our case I guess your approach is better

build-and-deploy-cloud-functions.sh
```diff
@@ -0,0 +1,33 @@
...
+echo "Deploying covid order registrar to ${PROJECT_ID}"
+gcloud --project=${PROJECT_ID} functions deploy \
+    order-in-care-evolve \
+    --entry-point=org.broadinstitute.dsm.cf.Covid19OrderRegistrarFunction \
+    --runtime=java11 \
+    --trigger-topic=tbos-ce-orders \
+    --source=target/deployment \
+    --memory=1024MB \
+    --set-env-vars="PROJECT_ID=${PROJECT_ID},SECRET_ID=study-manager-config" \
+    --vpc-connector=projects/broad-ddp-dev/locations/us-central1/connectors/appengine-default-connect
```
pegahtah, line 33: Looking into the tutorial, I can't totally figure out if we have to deploy them every time ?

yufengwng: You'll need to do a deploy anytime you want to change something, whether that's code, memory, env vars, or other settings. I've been working with Cloud Functions myself for some Pepper stuff and I have a similar `deploy.sh` script for those functions. Happy to chat to discuss my learnings.

SimoneMaiwald:
closing that one since the base branch is old now

PR #94, opened 2021-01-18
Changing functionality to send error notification to slack channel

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-5446

Used the same approach like it was before for sending error notification to email.

Added class SlackAppender:

- append method which handles to categorize errors
- sendSlackNotification method which handles to send passed error message to slack
- getSlackMessagePayload factory method which creates message payload as java object
- configure method used in DSMServer.setupErrorNotification to configure SlackAppender  with slack credentials from config file

To make functionality work we need to add slack configurations in main .conf file:
`"slack": 
{
        "hook": "<slack-hook-url>", 
        "channel": "<slack-channel-name>"
    }
`

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/log/SlackAppender.java
+++ b/src/main/java/org/broadinstitute/dsm/log/SlackAppender.java
@@ -17,6 +17,7 @@ import org.apache.commons.logging.Log;
 import org.apache.http.HttpHeaders;
 import org.apache.http.entity.ContentType;
 import org.apache.log4j.AppenderSkeleton;
+import org.apache.log4j.Level;
 import org.apache.log4j.spi.LoggingEvent;
 import org.broadinstitute.ddp.util.Utility;
 import org.broadinstitute.dsm.DSMServer;
@@ -24,11 +25,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class SlackAppender extends AppenderSkeleton {
+
+
     public SlackAppender() {
     }
 
     private static final Logger logger = LoggerFactory.getLogger(SlackAppender.class);
-    public static final String APPENDER_NAME = "slackAppender";
     private static HttpClient httpClient;
     private static String appEnv;
     private static String schedulerName;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/dsm/log/SlackAppender.java
```diff
@@ -0,0 +1,144 @@
...
+                    + "\n" + ExceptionUtils.getStackTrace(e));
+        }
+    }
+
+    public SlackMessagePayload getSlackMessagePayload(long currentEpoch, String note, String messageType) {
+        SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+        String date = formatter.format(new Date(currentEpoch * 1000L));
+        SlackMessagePayload payload = new SlackMessagePayload(String.format("An error has been detected for *%s*." +
+                " Please go check the backend logs around *%s UTC*. \n" + "%s \n %s",
+                appEnv, date, messageType, note), slackChannel, "DSM", ":nerd_face:");
```
SimoneMaiwald: I think we should change DSM to "Study-Manager" 
I like the nerd_face ☺️

src/main/java/org/broadinstitute/dsm/log/SlackAppender.java
```diff
@@ -0,0 +1,144 @@
...
+            httpClient = HttpClient.newHttpClient();
+            SlackAppender.slackChannel = slackChannel;
+            schedulerName = scheduler;
+            configured = true;
+        } else {
+            throw new RuntimeException("Configure has already been called for this appender.");
+        }
+    }
+
+    static class SlackMessagePayload {
```
SimoneMaiwald, line 120: I think I would prefer the SlackMessagePayload to be an extra class

gmakharat: > I think I would prefer the SlackMessagePayload to be an extra class

Since this class is only used in SlackAppender, we created it as inner class, do you think that maybe later we will use this class somewhere else as well?

SimoneMaiwald: ok, leave it for now and if we need it somewhere else we can pull it out then

SimoneMaiwald (approved):
change "DSM" to "Study-Manager" in the SlackMessagePayload, then good to merge

PR #95, opened 2021-01-22
DDP-5532_filterChanges

pegahtah:
The JSON_ARRAY type has been added to the automata so  we can search the `test_result` and also save and reapply the filters

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
+++ b/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
@@ -10,19 +10,17 @@ import org.broadinstitute.dsm.db.structure.ColumnName;
 import org.broadinstitute.dsm.db.structure.DBElement;
 import org.broadinstitute.dsm.db.structure.TableName;
 import org.broadinstitute.dsm.exception.DuplicateException;
-import org.broadinstitute.dsm.model.*;
+import org.broadinstitute.dsm.model.Filter;
+import org.broadinstitute.dsm.model.NameValue;
+import org.broadinstitute.dsm.model.ParticipantColumn;
+import org.broadinstitute.dsm.model.TissueList;
 import org.broadinstitute.dsm.statics.DBConstants;
 import org.broadinstitute.dsm.util.PatchUtil;
 import org.broadinstitute.dsm.util.SystemUtil;
-import org.json.JSONArray;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -52,17 +52,18 @@
...
+            "                                    and kit.test_result is null and kit.CE_order is null " +
+            "                                    and (kit.kit_label is null or kit.tracking_return_id is null or kit.tracking_to_id is null)" +
+            ") " +
+            "                                     AS request" +
+            "                                    LEFT JOIN ddp_participant_exit ex ON (ex.ddp_instance_id = request.ddp_instance_id AND ex.ddp_participant_id = request.ddp_participant_id)" +
+            "                                    LEFT JOIN (SELECT subK.kit_type_id, subK.external_name from ddp_kit_request_settings dkc " +
+            "                                    LEFT JOIN sub_kits_settings subK ON (subK.ddp_kit_request_settings_id = dkc.ddp_kit_request_settings_id)) as subkits ON (subkits.kit_type_id = request.kit_type_id) " +
+            "                                WHERE ex.ddp_participant_exit_id is null AND request.ddp_instance_id = ? AND (external_order_status not like '%CANCELLED%' or external_order_status is null) AND external_order_number IS NOT NULL " +
```
pegahtah, line 66: I just removed the `\n` and put the two `))` in 2 lines, it is easier for running. fixes for test boston this way.

src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -714,7 +728,7 @@ else if (word.equals(Filter.CLOSE_PARENTHESIS)) {// end of selected options
                             }
                             break;
                         case 17:// in additional fields query find the table and the column
-                            names = word.split("\\.");
+                            String[] names = word.split("\\.");
```
SimoneMaiwald: why not add "if (word.contains(".")) {" here as well?

SimoneMaiwald: make a method and reuse, please (case 2 and case 17, case 24
only state is different

SimoneMaiwald (approved):

SimoneMaiwald (approved):

PR #103, opened 2021-02-10
DDP-5633 Retry UPS queries a few times before failing

zyme:
The UPS tracking endpoint is a bit flakey.  This adds a few retry attempts before erroring out.  I've had these changes on a local branch and have been using them as part of manual kit report generation--they've definitely made things more robust, especially for long running jobs.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/shipping/UPSTracker.java
+++ b/src/main/java/org/broadinstitute/dsm/shipping/UPSTracker.java
@@ -1,6 +1,5 @@
 package org.broadinstitute.dsm.shipping;
 
-import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsm/shipping/UPSTracker.java
```diff
@@ -43,12 +42,26 @@ public UPSTrackingResponse lookupTrackingInfo(String trackingId) {
...
+               try {
+                   Thread.sleep(5 * 1000);
+               } catch (InterruptedException interrupted) {
+                   logger.warn("Error while waiting for UPS retry", interrupted);
+               }
+            }
         }
-        catch (IOException e) {
-            throw new RuntimeException("couldn't get response from ups tracking of package " + trackingId, e);
+        if (error != null) {
```
pegahtah, line 62: Shouldn’t `error` get null every time we enter the `for` loop? So that we don't throw an error if we had one in the first try but was successful in the 2nd?

SimoneMaiwald: it does. L46

SimoneMaiwald (approved):

src/main/java/org/broadinstitute/dsm/shipping/UPSTracker.java
pegahtah: That's before the loop though

zyme: The early return on L50 will ignore any exceptions from previous errors.  If we fail 5 times, L63 will throw the last error from the latest attempt.  Does that sound about right?

SimoneMaiwald: yes

pegahtah: Yes that's right. I missed the `return` part. Thanks for the explanation @zyme

pegahtah (approved):

PR #104, opened 2021-02-12
DDP-5235 Edit participant test

Irakli-Khaladze:
Here is a test for whole functionality of edit participant feature, I also needed to add functionality to delete message in EditParticipantMessage. I also changed getters and setters with Lombok Data. I already had that change, but hadn't pushed yet.

Also I've added add and delete user methods in DBTestUtil, but as I was able to use test user I didn't use those methods. If you'd like I will delete them.
To run test we need to add "unitTesterEmail" in test-config  to find the userId of a test user and we need to add "dss_to_dsm_subscription" and "dsm_to_dss_topic" which will have name of the test topic and it's subscription.

p.s I checked out to this branch from the branch from previous PR, because I had already done most of the part in that branch, just hadn't pushed. I guess on previous pull request you used quash merge, probably that's why it shows so many commits.


src/test/java/org/broadinstitute/dsm/EditParticipantTest.java
```diff
@@ -0,0 +1,114 @@
...
+        userId = User.getUser(cfg.getString(UNIT_TESTER_EMAIL)).getUserId();
+    }
+
+    @Test
+    public void testEditParticipantFeature() {
+
+        String realm = null;
+
+        if (UserUtil.checkUserAccess(realm, Integer.toString(userId), "mr_view") ||
+                UserUtil.checkUserAccess(realm, Integer.toString(userId), "pt_list_view")) {
```
SimoneMaiwald: can you change that to the new access role `participant_edit`

src/main/java/org/broadinstitute/dsm/db/EditParticipantMessage.java
```diff
@@ -226,4 +191,25 @@ public static void updateMessageStatusById(@NonNull int messageId, @NonNull Stri
         }
     }
 
+    public static void deleteMessage(int userId) {
```
SimoneMaiwald: if that method is only used in the unit test, I would prefer moving it to the test class

src/test/java/org/broadinstitute/dsm/EditParticipantTest.java
Irakli-Khaladze: Sure, but one thing is that we will need to give permission participant_edit to unit tester user to make test pass

SimoneMaiwald (approved):

PR #106, opened 2021-02-16
DDP-5441 | kit upload check

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-5441

**Things to pay attention:** 
added method userExistsInRealm to check, if participant presented in csv document, belongs to the study on which this kit was uploaded, if so, then it checks, if participantId entered in the csv document matches to the participantId taken from the user which was checked in study's elastic search

**UPDATE**
According to ticket DDP-5440, removed participantId check, instead, added check for full name, if user exists in the study it checks that user's full name taken from document against full name taken from elastic search.

Some refactoring has done


src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
```diff
@@ -343,74 +350,102 @@ private void addKitRequest(Connection conn, String kitTypeName, KitRequestSettin
...
+    private boolean userExistsInRealm(String shortId, String realm, String participantFirstNameFromDoc,
+                                      String participantLastNameFromDoc) {
+        if (StringUtils.isBlank(shortId) && shortId.matches("^[a-zA-Z0-9]*$")) {
+            return false;
+        }
+        if (participantFirstNameFromDoc == null && participantLastNameFromDoc == null) return false;
+
+        Map<String, String> queryConditions = new HashMap<>();
+        queryConditions.put("ES", " AND profile.hruid = '" + shortId + "'");
+        DDPInstance ddpInstanceByRealm = DDPInstance.getDDPInstance(realm);
```
SimoneMaiwald: can be moved further out and only the DDPInstance given into the method, because DDPInstance will always be the same for all pts in the upload so no need to get the instance for every participant

SimoneMaiwald (approved):

PR #107, opened 2021-02-16
* adding a check for the `test_result` not being null and for duplicate DSS notifications

pegahtah:
*adding a check for whether or not we have already triggered a notification to DSS before, if so, it will throw an error of kind " Kit DDP Notification was null for kit " so we can check this in the logs.


src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -45,6 +46,8 @@
     private static final String SQL_SELECT_KIT_FOR_NOTIFICATION_EXTERNAL_SHIPPER = "select  eve.*,   request.ddp_participant_id,   request.ddp_label,   request.dsm_kit_request_id, request.ddp_kit_request_id, request.upload_reason, " +
             "        realm.ddp_instance_id, realm.instance_name, realm.base_url, realm.auth0_token, realm.notification_recipients, realm.migrated_ddp, kit.receive_date, kit.scan_date" +
             "        from ddp_kit_request request, ddp_kit kit, event_type eve, ddp_instance realm where request.dsm_kit_request_id = kit.dsm_kit_request_id and request.ddp_instance_id = realm.ddp_instance_id" +
+            "        and kit.dsm_kit_request_id not in (select eq.dsm_kit_request_id  FROM event_type et left join EVENT_QUEUE eq on (et.ddp_instance_id=eq.ddp_instance_id) " +
```
zyme: Let's try a slightly different approach, where we to a `not exists` and join to tables that are already part of the main query:

```


select  eve.*,   request.ddp_participant_id,   request.ddp_label,   request.dsm_kit_request_id, request.ddp_kit_request_id, request.upload_reason,
                    realm.ddp_instance_id, realm.instance_name, realm.base_url, realm.auth0_token, realm.notification_recipients, realm.migrated_ddp, kit.receive_date, kit.scan_date
                    from ddp_kit_request request, ddp_kit kit, event_type eve, ddp_instance realm where request.dsm_kit_request_id = kit.dsm_kit_request_id and request.ddp_instance_id = realm.ddp_instance_id
                    and (eve.ddp_instance_id = request.ddp_instance_id and eve.kit_type_id = request.kit_type_id) and eve.event_type = ?

                    and
                    not exists(select 1 from EVENT_QUEUE q
                    where q.DDP_INSTANCE_ID = realm.ddp_instance_id
                    and
                    q.EVENT_TYPE = eve.event_type
                    and
                    q.DSM_KIT_REQUEST_ID = request.dsm_kit_request_id
                    )
```

This approach doesn't require additional predicates or params.

However, this query is quite slow.  It looks like `event_queue` should have an index on `event_type`, and `dsm_kit_request_id` should have a foreign key back to `dsm_kit_request`?

zyme: Can we also set `ddp_instance_id` somewhere in here and restrict to `event_triggered = true`?

zyme: `event_type.event_type` and  `event_type.event_name` should also be indexed...

SimoneMaiwald:
@pegahtah did you make a new PR for that change?

PR #112, opened 2021-02-19
spring cleaning 

SimoneMaiwald:
- removed old unused code
- removed old db updates (from before liquidbase)
- refactored because of unused role
- refactored because of all studies having participant_index


SimoneMaiwald:
More of an FYI =)

PR #117, opened 2021-02-24
Ddp 5445 ups cf

pegahtah:
ready for review


build-and-deploy-cloud-functions.sh
```diff
@@ -0,0 +1,45 @@
+PROJECT_ID=$1
+STUDY_MANAGER_SCHEMA=$2
+STUDY_SERVER_SCHEMA=$3
+
+echo "Will deploy to ${PROJECT_ID} using schemas ${STUDY_MANAGER_SCHEMA} and ${STUDY_SERVER_SCHEMA}"
+
+ mvn -Pcloud-function -DskipTests clean install package
+
+
+echo "Deploying kit dispatcher to ${PROJECT_ID}"
```
SimoneMaiwald, line 10: can we make that deploy script generic or do we always want to re-deploy all cfs?

src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -594,39 +594,12 @@ private void setupJobs(@NonNull Config cfg, @NonNull KitUtil kitUtil,
                 createScheduleJob(scheduler, null, null, EasypostShipmentStatusJob.class, "CHECK_STATUS_SHIPMENT",
                         cfg.getString(ApplicationConfigConstants.QUARTZ_CRON_STATUS_SHIPMENT), new EasypostShipmentStatusTriggerListener(), cfg);
 
-                //pegah todo
 
                 UPS_ACCESSKEY = cfg.getString(UPS_PATH_TO_ACCESSKEY);
                 UPS_USERNAME = cfg.getString(UPS_PATH_TO_USERNAME);
                 UPS_PASSWORD = cfg.getString(UPS_PATH_TO_PASSWORD);
                 UPS_ENDPOINT = cfg.getString(UPS_PATH_TO_ENDPOINT);
```
SimoneMaiwald: do we need them here?

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -60,184 +66,78 @@ public Covid19OrderRegistrar(String endpoint,
...
         if (kitPickupTime == null) {
             throw new CareEvolveException("Cannot place order for " + kitLabel + " without a pickup time");
         }
 
-        DDPInstance instance = DDPInstance.getDDPInstanceWithRole("testboston", DBConstants.HAS_KIT_REQUEST_ENDPOINTS);
-
-        Map<String, String> queryConditions = new HashMap<>();
-        queryConditions.put("ES", " AND profile.hruid = '" + participantHruid +"'");
-        List<ParticipantWrapper> participants = ParticipantWrapper.getFilteredList(instance, queryConditions);
+        DDPInstance ddpInstance = DDPInstance.getDDPInstanceWithRole("testboston", DBConstants.HAS_KIT_REQUEST_ENDPOINTS);
```
SimoneMaiwald: is there a reason why we check if the instance has that role? @zyme

src/main/java/org/broadinstitute/dsm/cf/Covid19OrderRegistrarFunction.java
```diff
@@ -0,0 +1,152 @@
...
+        Provider provider = new Provider(cfg.getString(ApplicationConfigConstants.CARE_EVOLVE_PROVIDER_FIRSTNAME),
+                cfg.getString(ApplicationConfigConstants.CARE_EVOLVE_PROVIDER_LAST_NAME),
+                cfg.getString(ApplicationConfigConstants.CARE_EVOLVE_PROVIDER_NPI));
+
+        PoolingDataSource<PoolableConnection> dataSource = CFUtil.createDataSource(5, cfg.getString(DSM_DB_URL));
+
+        DDPInstance ddpInstance = null;
+
+        try (Connection conn = dataSource.getConnection()) {
+            ddpInstance = DDPInstance.getDDPInstanceWithRole(conn,"testboston", DBConstants.HAS_KIT_REQUEST_ENDPOINTS);
```
SimoneMaiwald: is there a reason for checking that role? this is the role for checking dss for kits every hour, this is also the role we sometimes remove if we want to stop dsm from asking for new kits! if we need to check if the study does kit shipping or not `kit_request_activated` would be the safer choice! @zyme

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+                " req.external_order_number, kit.CE_order FROM " + STUDY_MANAGER_SCHEMA + "ddp_kit kit LEFT JOIN " + STUDY_MANAGER_SCHEMA + "ddp_kit_request req " +
+                " ON (kit.dsm_kit_request_id = req.dsm_kit_request_id) " +
+                "left join " + STUDY_MANAGER_SCHEMA + "kit_tracking_history kth on ( kth.dsm_kit_request_id = kit.dsm_kit_request_id) " +
+                " WHERE req.ddp_instance_id = ? and kit_label not like \"%\\\\_1\"  ";
+        String SQL_AVOID_DELIVERED = " and (tracking_to_id is not null or tracking_return_id is not null ) and (kth.kit_shipping_history is null or kth.kit_return_history is null or kth.kit_shipping_history not like \"" + UPSStatus.DELIVERED_TYPE + " %\" or kth.kit_return_history not like \"" + UPSStatus.DELIVERED_TYPE + " %\")" +
+                " order by kit.dsm_kit_request_id ASC";
+
+        logger.info("Starting the UPS lookup job");
+        LOOKUP_CHUNK_SIZE = cfg.getInt("kitTracking.trackingPerCFSize"); // todo pegah put in constants
```
SimoneMaiwald: I agree with the todo

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+        ArrayList<DdpKit> subsetOfKits = new ArrayList<>();
+        // static vars are dangerous in CF https://cloud.google.com/functions/docs/bestpractices/tips#functions-tips-scopes-java
+
+        try (Connection conn = dataSource.getConnection()) {
+            List<DDPInstance> ddpInstanceList = getDDPInstanceListWithRole(conn, "ups_tracking");
```
SimoneMaiwald: move "ups_tracking" to DBConstants

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+                    }
+                    conn.close();
+                }
+
+
+
+            }
+            String json = new Gson().toJson(subsetOfKits.toArray(), DdpKit[].class);
+            logger.info(json);
+            kitTrackerPubSubPublisher.publishWithErrorHandlerExample("broad-ddp-dev", "cf-kit-tracking", json);
```
SimoneMaiwald: needs to be changed ("broad-ddp-dev" and "cf-kit-tracking")

src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
```diff
@@ -60,12 +71,14 @@
...
     private final String usersIndexES;
+    private final String carrierTrackingUrl;
 
     private InstanceSettings instanceSettings;
 
     public DDPInstance(String ddpInstanceId, String name, String baseUrl, String collaboratorIdPrefix, boolean hasRole,
                        int daysMrAttentionNeeded, int daysTissueAttentionNeeded, boolean hasAuth0Token, List<String> notificationRecipient,
-                       boolean migratedDDP, String billingReference, String participantIndexES, String activityDefinitionIndexES, String usersIndexES) {
+                       boolean migratedDDP, String billingReference, String participantIndexES, String activityDefinitionIndexES,
+                     String carrierTrackingUrl, String usersIndexES) {
```
SimoneMaiwald: when changing the constructor can you please add it at the end

src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
```diff
@@ -79,6 +92,7 @@ public DDPInstance(String ddpInstanceId, String name, String baseUrl, String col
         this.billingReference = billingReference;
         this.participantIndexES = participantIndexES;
         this.activityDefinitionIndexES = activityDefinitionIndexES;
+        this.carrierTrackingUrl = carrierTrackingUrl;
```
SimoneMaiwald: move to the end as well

src/main/java/org/broadinstitute/dsm/db/DdpKit.java
```diff
@@ -24,13 +27,17 @@
...
+
     private static final Logger logger = LoggerFactory.getLogger(DdpKit.class);
 
     public DdpKit(String DsmKitRequestId, String kitLabel, String trackingToId, String trackingReturnId, String error,
-                  String message, String receiveDate, String upsTrackingStatus, String upsTrackingDate,
-                  String upsReturnStatus, String upsReturnDate, String bspCollaboratodId, String externalOrderNumber,
-                  boolean CEOrdered) {
+                  String message, String receiveDate, String bspCollaboratodId, String externalOrderNumber,
+                  boolean CEOrdered, String kitShippingHistory, String kitReturnHistory, String ddpInstanceId,
+                  String upsTrackingStatus, String upsTrackingDate, String upsReturnStatus, String upsReturnDate) {
```
SimoneMaiwald, line 40: should they still be here?

src/main/java/org/broadinstitute/dsm/db/DdpKit.java
```diff
@@ -60,7 +70,46 @@ public void changeCEOrdered(Connection conn, boolean orderStatus) {
                     + " to " + orderStatus);
         }
         catch (Exception e) {
-            throw new RuntimeException("Could not update ce_ordered status for " + this.getDsmKitRequestId(),e);
+            throw new RuntimeException("Could not update ce_ordered status for " + this.getDsmKitRequestId(), e);
+        }
+    }
+
+    public static boolean hasKitBeenOrderedInCE(Connection conn, String kitLabel) {
+        String query = "select k.ce_order from ddp_kit k where k.kit_label = ?";
```
SimoneMaiwald, line 57: can you move the queries out, so that someone looking to find queries doesn't need to go through the whole class and search for them

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,202 +1,211 @@
...
         }
     }
 
-    public static UPSTrackingResponse lookupTrackingInfo(String trackingId) {
-        return new UPSTracker(DSMServer.UPS_ENDPOINT, DSMServer.UPS_USERNAME, DSMServer.UPS_PASSWORD, DSMServer.UPS_ACCESSKEY).lookupTrackingInfo(trackingId);
+    public UPSTrackingResponse lookupTrackingInfo(Config cfg, String trackingId) throws IOException {
+        String endpoint = cfg.getString("ups.url");
+        String username = cfg.getString("ups.username");
+        String password = cfg.getString("ups.password");
+        String accessKey = cfg.getString("ups.accesskey");
```
SimoneMaiwald: move the strings out of the method

zyme:
We should probably revert all the CareEvolve changes--they were part of the POC for CFs, but I don't think we want them going forward.

pom.xml
```diff
@@ -234,54 +230,190 @@
...
-                </configuration>
-            </plugin>
-        </plugins>
+            <sourceDirectory>src/main/java</sourceDirectory>
+            <testSourceDirectory>src/test/java</testSourceDirectory>
     </build>
-</project>
 
-<!-- arz remove ruby, postgress, activejdbc -->
+<!--    <build>-->
```
zyme: Let's just delete these lines rather than commenting them out.

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Identifies kits whose shipping history should be
+ * queried (by another job) in order to produce
+ * reports on turnaround time and study adherence.
+ */
+public class TestBostonKitTrackerDispatcher implements BackgroundFunction<PubsubMessage> {
+
+    private static final Logger logger = Logger.getLogger(TestBostonKitTrackerDispatcher.class.getName());
+
+    private String STUDY_MANAGER_SCHEMA = System.getenv("STUDY_MANAGER_SCHEMA") + ".";
```
zyme, line 43: In the original POC, I had to run a query across schemas, but here I think we can set the URL params for the db to use the appropriate DSM db.  That will save some ickiness with SQL string concatenation.

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+    private int LOOKUP_CHUNK_SIZE;
+
+
+    KitTrackerPubSubPublisher kitTrackerPubSubPublisher = new KitTrackerPubSubPublisher();
+
```
zyme, line 46: Let's try to save some vertical space so we can view more code at once and just separate things with a single vertical line instead of two.

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
zyme: A bind parameter (`?`) sounds good, and then this can be passed into whatever is running the query.

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+                " ON (kit.dsm_kit_request_id = req.dsm_kit_request_id) " +
+                "left join " + STUDY_MANAGER_SCHEMA + "kit_tracking_history kth on ( kth.dsm_kit_request_id = kit.dsm_kit_request_id) " +
+                " WHERE req.ddp_instance_id = ? and kit_label not like \"%\\\\_1\"  ";
+        String SQL_AVOID_DELIVERED = " and (tracking_to_id is not null or tracking_return_id is not null ) and (kth.kit_shipping_history is null or kth.kit_return_history is null or kth.kit_shipping_history not like \"" + UPSStatus.DELIVERED_TYPE + " %\" or kth.kit_return_history not like \"" + UPSStatus.DELIVERED_TYPE + " %\")" +
+                " order by kit.dsm_kit_request_id ASC";
+
+        logger.info("Starting the UPS lookup job");
+        LOOKUP_CHUNK_SIZE = cfg.getInt("kitTracking.trackingPerCFSize"); // todo pegah put in constants
+        ArrayList<DdpKit> subsetOfKits = new ArrayList<>();
+        // static vars are dangerous in CF https://cloud.google.com/functions/docs/bestpractices/tips#functions-tips-scopes-java
```
zyme: This comment is interesting, but I don't think it belongs here?

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+
+                            while (rs.next() && count < LOOKUP_CHUNK_SIZE) {
+                                count++;
+                                String kitLabel = rs.getString("kit.kit_label");
+                                logger.info(kitLabel);
+
+                                DdpKit kit = new DdpKit(
+                                        rs.getString(DBConstants.DSM_KIT_REQUEST_ID),
+                                        rs.getString("kit." + DBConstants.KIT_LABEL),
+                                        rs.getString("kit." + DBConstants.DSM_TRACKING_TO),
```
zyme: Let's move these constants (`"kit."`, `"req."`) into constants

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+                                        rs.getString("kit." + DBConstants.UPS_RETURN_STATUS),
+                                        rs.getString("kit." + DBConstants.UPS_RETURN_DATE)
+                                );
+                                subsetOfKits.add(kit);
+                                logger.info("added " + kitLabel);
+
+                            }
+                        }
+                    }
+                    conn.close();
```
zyme: `conn` will be autoclosed by virtue of it being declared within the `try()`, so no need to close it here.

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
zyme: These values should be parameterized in some way.  I believe they are being passed through as CF environment variables, so let's read them from the environment.

src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
```diff
@@ -267,7 +276,7 @@ public static boolean getRole(@NonNull String realm, @NonNull String role) {
         return (boolean) results.resultValue;
     }
 
-    private static DDPInstance getDDPInstanceWithRoleFormResultSet(@NonNull ResultSet rs) throws SQLException {
+    public static DDPInstance getDDPInstanceWithRoleFormResultSet(@NonNull ResultSet rs) throws SQLException {
```
zyme, line 360: Passing in a `ResultSet` makes it hard to see what the method is really operating on (and can also make things hard to test).  Can you pull out the values you need from the resultset and create arguments for each?

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,202 +1,211 @@
...
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.List;
 
-import static org.broadinstitute.ddp.db.TransactionWrapper.inTransaction;
 
-public class TestBostonUPSTrackingJob implements Job {
+public class TestBostonUPSTrackingJob implements BackgroundFunction<PubsubMessage> {
```
zyme, line 32: I think we can make the object passed in here much more specific.  `PubsubMessage` is from the POC, and it's a bit of a mistake that I added it here.  See if you can parameterize this class with `DdpKit[]` or a variation thereupon so that you don't have to parse `data`.  The framework should do this for you.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,202 +1,211 @@
...
+            return;
+        }
+
+
+        logger.info(message.data + " *");
+        Config cfg = CFUtil.loadConfig();
+
+        String dbUrl = cfg.getString("dsmDBUrl");
+
+        PoolingDataSource<PoolableConnection> dataSource = CFUtil.createDataSource(20, dbUrl);
```
zyme: Do we need 20 connections in the pool?  If this is single threaded, maybe we only need one?  It's important because if we proliferate CFs that each create so many connections, eventually we'll run out of them on the server side.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,202 +1,211 @@
...
         String errorString = "";
         for (UPSError error : errors) {
             errorString += "Got Error: " + error.getCode() + " " + error.getMessage() + " For Tracking Number " + trackingId;
         }
         logger.error(errorString);
     }
 
-    private static void updateStatus(@NonNull Connection conn, String trackingId, String oldType, UPSTrackingResponse response, boolean isReturn, DdpKit kit, DDPInstance ddpInstance) {
+
+    private void updateStatus(PoolingDataSource<PoolableConnection> dataSource, String trackingId, UPSActivity lastActivity, UPSTrackingResponse response,
```
zyme: This should all be transactional, right?  If so, passing around the `dataSource` will potentially make this non-transactional, because each call to `getConnection()` can return a *different connection*, which may result in very weird bugs.  Please use a single connection instead of passing through the `dataSource`.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -205,77 +214,121 @@ private static void updateStatus(@NonNull Connection conn, String trackingId, St
...
+        Connection conn = null;
+        try {
+            conn = dataSource.getConnection();
+            try (PreparedStatement stmt = conn.prepareStatement(query)) {
+                stmt.setString(1, kit.getDsmKitRequestId());
+                stmt.setString(2, upsHistory);
+                stmt.setString(3, upsHistory);
+                int r = stmt.executeUpdate();
+                conn.commit();
+                logger.info("Updated " + r + " rows");
```
zyme: I'm not sure the reader of the logs will know what's going on without more context-- `Updated 3 rows` is somewhat ambiguous...

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -205,77 +214,121 @@ private static void updateStatus(@NonNull Connection conn, String trackingId, St
...
-            throw new RuntimeException("Could not update tracking info for tracking id " + trackingId, e);
+            throw new RuntimeException("unable to connect to DB " + e);
+        }
+        finally {
+            if (conn != null) {
+                try {
+                    conn.close();
+                }
+                catch (Throwable ex) {
+                    logger.debug("Could not close JDBC Connection on shutdown", ex);
```
zyme: This should be an `error`--failure to close is interesting enough to warrant it.  Also, "shutdown" isn't what's going on here.  The method is exiting normally.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -205,77 +214,121 @@ private static void updateStatus(@NonNull Connection conn, String trackingId, St
...
+
+                        }
+                        else {
+                            logger.error("received kitDDPNotification was null for " + kit.getExternalOrderNumber());
+                        }
                     }
                 }
+                logger.info("Updated status of tracking number " + trackingId + " to " + statusType + " from " + oldType + " for kit w/ external order number " + kit.getExternalOrderNumber());
+            }
+            catch (SQLException ex) {
```
zyme, line 468: When `Connection` is passed in here instead of `dataSource`, this exception handling streamlined.

Just as an FYI, when code gets wrapped in multiple nested `try` blocks, things get hard to understand.  Consider an alternative approach, where you can isolate errors a bit better rather than opening a new `try` on L244
```

try {
            conn = dataSource.getConnection();
        } catch (SQLException e) {
            throw new RuntimeException("Could not get a connection", e);
        }
```

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,202 +1,211 @@
...
 
-    public static UPSTrackingResponse lookupTrackingInfo(String trackingId) {
-        return new UPSTracker(DSMServer.UPS_ENDPOINT, DSMServer.UPS_USERNAME, DSMServer.UPS_PASSWORD, DSMServer.UPS_ACCESSKEY).lookupTrackingInfo(trackingId);
+    public UPSTrackingResponse lookupTrackingInfo(Config cfg, String trackingId) throws IOException {
+        String endpoint = cfg.getString("ups.url");
+        String username = cfg.getString("ups.username");
+        String password = cfg.getString("ups.password");
+        String accessKey = cfg.getString("ups.accesskey");
+        logger.info(endpoint);
+        return new UPSTracker(endpoint, username, password, accessKey).lookupTrackingInfo(trackingId);
```
zyme: How about declaring `UPSTracker` as a field?  Then these config values can be grabbed early in the lifecycle of the CF.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,202 +1,211 @@
...
+        String data = new String(Base64.getDecoder().decode(message.data));
+        logger.info("data " + data);
+
+        DdpKit[] kitsToLookFor = new Gson().fromJson(data, DdpKit[].class);
+        for (DdpKit kit : kitsToLookFor) {
+            if (StringUtils.isNotBlank(kit.getTrackingToId()) || StringUtils.isNotBlank(kit.getTrackingReturnId())) {
+                logger.info("Checking UPS status for kit with external order number " + kit.getExternalOrderNumber());
+                if (StringUtils.isNotBlank(kit.getTrackingToId()) && !kit.isDelivered()) {
+                    try {
+                        updateKitStatus(dataSource, kit, false, kit.getDdpInstanceId(), cfg);
```
zyme: Passing around `cfg` makes things hard to test--can you extract the vars you need at the top level, and the parameterize things accordingly?

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
zyme: Rather than having the chunk size set in the config file--which requires deployment--let's read it from the pubsub message itself that is passed into `accept()`.  This way, you can fiddle with it in whatever starts the job without having to redeploying the job.

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+                }
+
+
+
+            }
+            String json = new Gson().toJson(subsetOfKits.toArray(), DdpKit[].class);
```
zyme: Thinking back to some of the scale issues we saw earlier in GAE: serializing potentially large objects to Strings is something we should avoid.  Instead, use a more stream-oriented approach.  We should also avoid logging large objects like this on L112.

Consider passing `ArrayList<DdpKit>` instead of `json` to `kitTrackerPubSubPublisher.publishWithErrorHandlerExample`.  Then within that method, look for a more stream-oriented methods such as  `ByteString.readFrom()` and stream-oriented `Gson` methods.

src/main/java/org/broadinstitute/dsm/pubsub/KitTrackerPubSubPublisher.java
```diff
@@ -0,0 +1,57 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+
+
+public class KitTrackerPubSubPublisher {
+
+    private final Logger logger = LoggerFactory.getLogger(KitTrackerPubSubPublisher.class);
+    public void publishWithErrorHandlerExample(String projectId, String topicId, String message)
```
zyme: I'm guessing this was copy-pasted from somewhere?  How about renaming to `publishMessage()`?

src/main/java/org/broadinstitute/dsm/pubsub/KitTrackerPubSubPublisher.java
```diff
@@ -0,0 +1,57 @@
...
+            ByteString data = ByteString.copyFrom(message, StandardCharsets.UTF_8);
+            PubsubMessage pubsubMessage = PubsubMessage.newBuilder().setData(data).build();
+
+            // Once published, returns a server-assigned message id (unique within the topic)
+            Publisher publisher = Publisher.newBuilder(
+                    ProjectTopicName.of(projectId, topicId)).build();
+
+            String responseMessage;
+            try {
+                publisher.publish(pubsubMessage).get();
```
zyme: Probably better to avoid the blocking `get()` and call `addListener()`.  The pubsub message id that is ultimately returned here should be logged as well.  Check out some of the birch code to see some examples of this.

src/main/java/org/broadinstitute/dsm/pubsub/KitTrackerPubSubPublisher.java
```diff
@@ -0,0 +1,57 @@
...
+                responseMessage = "Message published. ";
+            } catch (InterruptedException | ExecutionException e) {
+                logger.error("Error publishing Pub/Sub message: " + e.getMessage(), e);
+                responseMessage = "Error publishing Pub/Sub message; see logs for more info. ";
+            }finally {
+                publisher.shutdown();
+            }
+
+
+        logger.info(responseMessage+message);
```
zyme: Probably too big a thing to log...

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -21,4 +21,23 @@
...
\ No newline at end of file
+
+    <changeSet id="carrier_per_study_key" author="pegah">
+        <addColumn tableName="ddp_instance">
+            <column name="carrier_tracking_url" type="VARCHAR(500)"></column>
+        </addColumn>
+    </changeSet>
+    <changeSet id="DDP-5445-tracking_history_table" author="pegah">
+        <createTable tableName="kit_tracking_history">
+            <column name="dsm_kit_request_id" type="INT(11)" >
```
zyme: Each table should have its own auto-generated primary key, rather than relying on the PK of another table.  Can you add the pk column and then change this to a foreign key on kit_request?

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -21,4 +21,23 @@
     </changeSet>
     <!-- prefix for gen2 prostate pts in dev: 1771332079 on prod: -763171247  -->
     <!-- prod i.instance_id > 20000 are actual new added initial_biopsies for migrated pts  -->
-</databaseChangeLog>
\ No newline at end of file
+
+    <changeSet id="carrier_per_study_key" author="pegah">
+        <addColumn tableName="ddp_instance">
+            <column name="carrier_tracking_url" type="VARCHAR(500)"></column>
```
zyme, line 27: This seems more like a `carrier_service` or `ddp_kit_request_settings` kind of thing to me.  Is `ddp_instance` the best place for this?

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -21,4 +21,23 @@
...
+        </addColumn>
+    </changeSet>
+    <changeSet id="DDP-5445-tracking_history_table" author="pegah">
```
zyme: Can we serialize this out into more structured tables that track dates and event types?  It will be challenging to deserialize everything as pojos and build performant queries.

zyme:
There are a few specifics in DDP-5445 that should be added here, particularly around:
* avoiding writing history when what we've read is less than what we had stored previously
* safeguards to detect and ignore kit history when UPS re-uses a label for a different package ~180 days after we used it
* configuration that maps UPS API credentials to particular studies

build-and-deploy-cloud-functions.sh
zyme: I'm in favor of always re-deploying so that we have something nice and simple: whenever one CF is updated, they're all updated.

src/main/java/org/broadinstitute/dsm/cf/CFUtil.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.commons.dbcp2.PoolingDataSource;
+import org.apache.commons.pool2.ObjectPool;
+import org.apache.commons.pool2.impl.GenericObjectPool;
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+
+public class CFUtil {
+
+    private static final Logger logger = Logger.getLogger(CFUtil.class.getName());
+
+    // todo centralize this for stateless db connections.  statics in cloud functions
```
zyme: I think this has been to-done, so the comment can be removed?

src/test/java/org/broadinstitute/dsm/GBFTest.java
```diff
@@ -225,4 +269,61 @@ public void inputConfirmationGBFKit() throws Exception {
...
+                            String ddpParticipantId = rs.getString("ddp_participant_id");
+                            DDPParticipant participant = ElasticSearchUtil.getParticipantAsDDPParticipant(elasticMap, ddpParticipantId);
+
+                            kitsToOrder.add(new KitRequest(ddpParticipantId, participant.getShortId(), participant, externalOrderNumber));
+
+                            for (KitRequestSettings kitRequestSetting : kitRequestSettings.values()) {
+                                if ("gbf".equalsIgnoreCase(kitRequestSetting.getExternalShipper())) {
+                                    kitRequestSetting.setCarrierTo("3rd Day Air Residential");
+                                    kitRequestSetting.setServiceTo(kitRequestSetting.getCarrierTo());
+                                    System.out.println(kitRequestSetting.getCarrierTo());
```
zyme: Please use a log statement instead of println.

src/test/java/org/broadinstitute/dsm/GBFTest.java
```diff
@@ -225,4 +269,61 @@ public void inputConfirmationGBFKit() throws Exception {
...
+                            }
+
+                        }
+                    }
+                }
+            } catch (Exception e) {
+                logger.error("Could not order kit " + externalOrderNumber, e);
+            }
+
+            return null;
```
zyme, line 309: I don't see any assertions here--can you add some tests with assertions?

src/main/java/org/broadinstitute/dsm/pubsub/KitTrackerPubSubPublisher.java
pegahtah: Just here for until the testing is done...

src/test/java/org/broadinstitute/dsm/GBFTest.java
pegahtah: This was made to be used to order kits through gbf, not as a real test. Since the new PR I don't think we will need it anymore...

src/main/java/org/broadinstitute/dsm/db/DdpKit.java
pegahtah: The CF tutorials on the web said not to do that, I'll find a link to it

src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
pegahtah: Agreed! I'll fix this

src/main/java/org/broadinstitute/dsm/model/UPSKit.java
```diff
@@ -0,0 +1,70 @@
...
+    Boolean CE_order;
+    String dsmKitRequestId;
+    String externalOrderNumber;
+    String trackingToId;
+    String trackingReturnId;
+    String ddpInstanceId;
+    String hruid;
+
+    private static final Logger logger = LoggerFactory.getLogger(UPSKit.class.getName());
+
```
pegahtah: I made a simpler and smaller object than `DdpKit` to use with the new UPS classes

src/main/resources/liquibase/042-updateDB.xml
pegahtah: moved to `carrier_service`

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -24,124 +21,186 @@
...
-
+        endpoint = cfg.getString("ups.url");
+        username = cfg.getString("ups.username");
+        password = cfg.getString("ups.password");
+        accessKey = cfg.getString("ups.accesskey");
+        PoolingDataSource<PoolableConnection> dataSource = CFUtil.createDataSource(1, dbUrl);
         logger.info("Starting the UPS lookup job");
 
+        try {
+            conn = dataSource.getConnection();
```
zyme: If you put `conn` into a try-with-resources, it'll autoclose for you and you can remove the `finally` on L85.

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -24,124 +21,186 @@
...
+            Arrays.stream(kitsToLookFor).forEach(
+                    kit -> {
+                        logger.info("Checking possible actions for kit " + kit.getDsmKitRequestId());
+                        if (StringUtils.isNotBlank(kit.getUpsPackage().getUpsShipmentId())) {
+                            getUPSUpdate(conn, kit, cfg);
+                        }
+                        else {
+                            insertShipmentAndPackageForNewKit(conn, kit, cfg);// for a new kit we first need to insert the UPSShipment
```
zyme: Given a potentially large volume of kits here and the flakiness of the UPS endpoint, I think it'd be be better to do the db updates for each kit in a separate connection/transaction.  In this way we'll commit updates as we make them and minimize the chance that one UPS query explosion will rollback all updates for all kits in the batch.

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -12,32 +12,74 @@
     </changeSet>
     <changeSet author="simone" id="setDeletedFlag">
         <sql>UPDATE ddp_medical_record as m, (
-                SELECT m.medical_record_id from ddp_medical_record m
-                left join ddp_institution i on (i.institution_id = m.institution_id)
-                left join ddp_participant p on (p.participant_id = i.participant_id)
-                left join ddp_instance realm on (p.ddp_instance_id = realm.ddp_instance_id)
-                where realm.instance_name = 'Prostate' and (p.ddp_participant_id like '1771332079%' OR p.ddp_participant_id like '-763171247%') and type = 'INITIAL_BIOPSY' and i.ddp_institution_id != '1' and i.institution_id &#60; 20000) as m2
-            SET m.deleted = 1 WHERE m2.medical_record_id = m.medical_record_id;</sql>
+            SELECT m.medical_record_id from ddp_medical_record m
```
zyme: This seems unrelated to ups kit tracking--should this change be in this branch?

src/main/java/org/broadinstitute/dsm/model/UPSKit.java
```diff
@@ -0,0 +1,70 @@
...
+        this.dsmKitRequestId = dsmKitRequestId;
+        this.externalOrderNumber = externalOrderNumber;
+        this.trackingToId = trackingToId;
+        this.trackingReturnId = trackingReturnId;
+        this.ddpInstanceId = ddpInstanceId;
+        this.hruid = hruid;
+    }
+
+    public boolean isReturn() {
+        return upsPackage.getTrackingNumber().equals(trackingReturnId);
```
zyme: This could NPE--can you make it more defensive?

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -12,13 +12,74 @@
...
+                <constraints nullable="false" primaryKey="true"></constraints>
+            </column>
+            <column name="dsm_kit_request_id" type="INT">
+                <constraints nullable="false"></constraints>
+            </column>
+            <column name="ups_shipment_id" type="INT">
+                <constraints nullable="false"></constraints>
+            </column>
+            <column name="tracking_number" type="varchar(50)"></column>
+            <column name="delivery_date" type="varchar(10)"></column><!-- Format: MMDDYYYY-->
```
zyme, line 55: Please make all the times actual `datetime(6)` to avoid app-level string parsing for date-based queries.

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -12,13 +12,74 @@
...
+                <constraints nullable="false"></constraints>
+            </column>
+            <column name="dsm_kit_request_id" type="INT">
+                <constraints nullable="false"></constraints>
+            </column>
+            <column name="ups_location" type="TEXT(10000)"></column>
+            <column name="ups_status_type" type="VARCHAR(10)"></column>
+            <column name="ups_status_description" type="VARCHAR(100)"></column>
+            <column name="ups_status_code" type="VARCHAR(10)"></column>
+            <column name="ups_activity_date" type="VARCHAR(10)"></column>
```
zyme: The date and time columns can be combined into a single `datetime` field for full precision and ease of querying.

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -12,13 +12,74 @@
...
+            <column name="delivery_time_type" type="varchar(3)"></column><!-- Indicates the type of delivery-->
+
+        </createTable>
+    </changeSet>
+    <changeSet id="DDP-5445_ups_activity_table" author="pegah">
+        <createTable tableName="ups_activity">
+            <column name="ups_activity_id" type="INT" autoIncrement="true">
+                <constraints nullable="false" primaryKey="true"></constraints>
+            </column>
+            <column name="ups_package_id" type="INT">
```
zyme, line 68: The various id references to other columns need to be marked as foreign keys.

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -12,13 +12,74 @@
...
+            <column name="ups_shipment_id" type="INT" autoIncrement="true">
+                <constraints nullable="false" primaryKey="true"></constraints>
+            </column>
+            <column name="dsm_kit_request_id" type="INT">
+                <constraints nullable="false"></constraints>
+            </column>
+        </createTable>
+
+    </changeSet>
+    <changeSet id="DDP-5445-ups_package-table" author="pegah">
```
zyme, line 44: Are there logical unique keys that can be added to each table to prevent application bugs such as repeatedly adding accidentally duplicated rows?

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -12,13 +12,74 @@
...
+            <column name="dsm_kit_request_id" type="INT">
+                <constraints nullable="false"></constraints>
+            </column>
+            <column name="ups_shipment_id" type="INT">
+                <constraints nullable="false"></constraints>
+            </column>
+            <column name="tracking_number" type="varchar(50)"></column>
+            <column name="delivery_date" type="varchar(10)"></column><!-- Format: MMDDYYYY-->
+            <column name="delivery_date_type" type="varchar(3)"></column><!-- Indicates the type of delivery-->
+            <column name="delivery_time_start" type="varchar(6)"></column><!-- Format: HHMMSS 24 hour-->
```
zyme, line 57: Are these delivery dates different from those one might find by querying `ups_activity` and looking for specific status types?  Or are these dates the projected delivery times from UPS?  Could you add a comment describing how these fields are different from the dates in `ups_activity`?

zyme:
Couple of open items that DDP-5445 mentions:
* how do we ensure that when UPS erases data, we don't erase our copy?
Screen Shot 2021-03-25 at 3 58 30 PM
* we need some verification that the shipping history is for our shipment.  Suggest hashing the address when the order is placed and only updating shipping records when the hash of the address in the UPS data matches our hash.

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
SimoneMaiwald: @pegahtah if the role is not used, can you please change it to just getDDPInstance

src/main/java/org/broadinstitute/dsm/careevolve/Patient.java
```diff
@@ -1,6 +1,23 @@
 package org.broadinstitute.dsm.careevolve;
 
+import static org.broadinstitute.dsm.careevolve.Covid19OrderRegistrar.ACTIVITIES_FIELD;
```
SimoneMaiwald: please remove unused imports

src/main/java/org/broadinstitute/dsm/cf/Covid19OrderRegistrarFunction.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import com.google.cloud.functions.BackgroundFunction;
+import com.google.cloud.functions.Context;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.google.gson.annotations.SerializedName;
+import com.google.pubsub.v1.PubsubMessage;
```
SimoneMaiwald: please remove unused import

src/main/java/org/broadinstitute/dsm/cf/Covid19OrderRegistrarFunction.java
SimoneMaiwald: @pegahtah if the role is not used, please change it to just getDDPInstance

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,192 @@
...
+                " ( SELECT ac.ups_activity_id  " +
+                " FROM ups_package pac INNER JOIN  " +
+                "    ( SELECT  ups_package_id, MAX(ups_activity_id) maxId  " +
+                "    FROM ups_activity  " +
+                "    GROUP BY ups_package_id  ) lastActivity ON pac.ups_package_id = lastActivity.ups_package_id INNER JOIN  " +
+                "    ups_activity ac ON   lastActivity.ups_package_id = ac.ups_package_id  " +
+                "    AND lastActivity.maxId = ac.ups_activity_id  " +
+                " ))";
+        String SQL_AVOID_DELIVERED = " and (tracking_to_id is not null or tracking_return_id is not null ) and  pack.delivery_date is null and kit.test_result is null " +
+                " order by kit.dsm_kit_request_id ASC LIMIT ?";// todo pegah
```
SimoneMaiwald: @pegahtah what is the todo for?

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,192 @@
...
+        logger.info("Starting the UPS lookup job");
+        logger.info(data);
```
SimoneMaiwald: remove or add more info, please

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,192 @@
...
+        LOOKUP_CHUNK_SIZE = new JsonParser().parse(data).getAsJsonObject().get("size").getAsInt();
+        logger.info("The chunk size for each cloud function is " + LOOKUP_CHUNK_SIZE);
+        JSONArray subsetOfKits = new JSONArray();
+        String project = cfg.getString("pubsub.projectId");
+        String topicId = cfg.getString("pubsub.topicId");
```
SimoneMaiwald: can we move these strings into `ApplicationConfigConstants`

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,192 @@
...
+                            stmt.setInt(3, LOOKUP_CHUNK_SIZE);
+                            subsetOfKits = new JSONArray();
+                            stmt.setFetchSize(LOOKUP_CHUNK_SIZE);
+                            try (ResultSet rs = stmt.executeQuery()) {
+                                while (rs.next()) {
+                                    logger.info(kit == null ? "null" : kit.getDsmKitRequestId());
+                                    String shipmentId = rs.getString("ups_shipment_id");
+                                    UPSPackage upsPackage;
+                                    if (StringUtils.isNotBlank(shipmentId)) {
+                                        UPSStatus latestStatus = new UPSStatus(rs.getString("activity.ups_status_type"),
```
SimoneMaiwald: can we move the column name strings into `DBConstants`

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,192 @@
...
+                                            rs.getString(DBConstants.DDP_KIT_TABLE_ABBR + DBConstants.DSM_TRACKING_TO),
+                                            rs.getString(DBConstants.DDP_KIT_TABLE_ABBR + DBConstants.DSM_TRACKING_RETURN),
+                                            rs.getString(DBConstants.DDP_KIT_REQUEST_TABLE_ABBR + DBConstants.DDP_INSTANCE_ID),
+                                            rs.getString(DBConstants.DDP_KIT_REQUEST_TABLE_ABBR + DBConstants.BSP_COLLABORATOR_PARTICIPANT_ID)
+                                    );
+                                    JSONObject jsonKit = new JSONObject(kit);
+                                    subsetOfKits.put(jsonKit);
+                                    logger.info("added " + kit.getKitLabel());
+                                    logger.info("size of array " + subsetOfKits.length());
+                                    logger.info("array " + subsetOfKits.toString());
```
SimoneMaiwald: make one logger.info out of this

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -0,0 +1,192 @@
...
+                                    kit = null;
+                                }
+                            }
+                            catch (Exception e) {
+                                logger.error("Trouble executing select query ", e);
```
SimoneMaiwald: do we want to add the kit id here to know where the problem was?

src/main/java/org/broadinstitute/dsm/db/DdpKit.java
```diff
@@ -45,52 +48,49 @@ public DdpKit(String DsmKitRequestId, String kitLabel, String trackingToId, Stri
...
-            stmt.setBoolean(1, orderStatus);
-            stmt.setString(2, this.getDsmKitRequestId());
-            int r = stmt.executeUpdate();
-            if (r != 1) {//number of subkits
-                throw new RuntimeException("Update query for CE order flag updated " + r + " rows! with dsm kit request id: " + this.getDsmKitRequestId());
+            stmt.setString(1, kitLabel);
+            ResultSet rs = stmt.executeQuery();
+
+            if (rs.next()) {
+                hasBeenOrdered = rs.getBoolean(1);
```
SimoneMaiwald, line 64: please use the column name here

src/main/java/org/broadinstitute/dsm/db/ParticipantEvent.java
```diff
@@ -133,5 +134,23 @@ public static void skipParticipantEvent(@NonNull String ddpParticipantId, @NonNu
...
+            stmt.setString(1, instanceId);
+            stmt.setString(2, ddpParticipantId);
+            try (ResultSet rs = stmt.executeQuery()) {
+                while (rs.next()) {
+                    skippedEvents.add(rs.getString(DBConstants.EVENT));
+                }
+            }
+        }
+        catch (Exception ex) {
+            logger.error("Couldn't exited participants for " + instanceId, ex);
```
SimoneMaiwald: please update the error msg

src/main/java/org/broadinstitute/dsm/db/ParticipantEvent.java
```diff
@@ -133,5 +134,23 @@ public static void skipParticipantEvent(@NonNull String ddpParticipantId, @NonNu
         }
         return skippedEvents;
     }
+
+    public static Collection<String> getParticipantEvent(Connection conn, @NonNull String ddpParticipantId, @NonNull String instanceId) {
+        ArrayList<String> skippedEvents = new ArrayList();
+        try (PreparedStatement stmt = conn.prepareStatement(TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.GET_PARTICIPANT_EVENT))) {
```
SimoneMaiwald: can you move sql for `GET_PARTICIPANT_EVENT` from the application.conf to the class, please

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
+    String username;
+    String password;
+    String accessKey;
+    private final int MAX_CONNECTION = 1;
 
     @Override
-    public void execute(JobExecutionContext context) throws JobExecutionException {
+    public void accept(PubsubMessage message, Context context) throws Exception {
+        if (message.data == null) {
+            logger.info("No message provided");
```
SimoneMaiwald: that sounds more like an error than an info

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
+            return;
+        }
+        Config cfg = CFUtil.loadConfig();
+        String dbUrl = cfg.getString("dsmDBUrl");
```
SimoneMaiwald, line 59: can we move the strings into `ApplicationConfigConstants`

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
+        PoolingDataSource<PoolableConnection> dataSource = CFUtil.createDataSource(MAX_CONNECTION, cfg.getString(ApplicationConfigConstants.CF_DSM_DB_URL));
+        try (Connection conn = dataSource.getConnection()) {
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_INSERT_SHIPMENT, Statement.RETURN_GENERATED_KEYS)) {
+                stmt.setString(1, kit.getDsmKitRequestId());
+                int result = stmt.executeUpdate();
+                if (result == 1) {
+                    try (ResultSet rs = stmt.getGeneratedKeys()) {
+
+                        if (rs.next()) {
+                            insertedShipmentId = rs.getString(1);
```
SimoneMaiwald, line 113: please us the column name

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
+                                insertedPackageIds[i] = rs.getString(1);
+                                i++;
+                            }
+                            if (i != 2) {
+                                throw new RuntimeException("Didn't insert right amount of packages. Num of Packages = " + i);
+                            }
+                            shippingKitPackageId = insertedPackageIds[0];
+                            returnKitPackageId = insertedPackageIds[1];
+                            logger.info("Added new ups package with id " + shippingKitPackageId + " for " + kit.getDsmKitRequestId());
+                            logger.info("Added new ups package with id " + returnKitPackageId + " for " + kit.getDsmKitRequestId());
```
SimoneMaiwald: can we make one logger.info out of this

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
+                        }
+                        catch (Exception e) {
+                            logger.error("Error getting id of new packages  ", e);
```
SimoneMaiwald: can you add the kit id or some id

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
-        if (isReturn) {
-            type = kit.getUpsReturnStatus();
-        }
-        else {
-            type = kit.getUpsTrackingStatus();
-        }
-        if (StringUtils.isNotBlank(type)) {// get only type from it
-            type = type.substring(0, type.indexOf(' '));
+    public void updateKitStatus(UPSKit kit, boolean isReturn, String ddpInstanceId, Config cfg) {
+        String trackingId = kit.getUpsPackage().getTrackingNumber();
```
SimoneMaiwald: add a check for kit.getUpsPackage() != null

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
+        //        System.out.println("Converted String: " + strDate);
+        for (int i = activities.length - 1; i >= 0; i--) {
+            UPSActivity currentInsertingActivity = activities[i];
+            if (lastActivity != null && lastActivity.getInstant() != null && (currentInsertingActivity.getInstant().equals(lastActivity.getInstant()) || currentInsertingActivity.getInstant().isBefore(lastActivity.getInstant()))) {
+                break;
             }
+            PoolingDataSource<PoolableConnection> dataSource = CFUtil.createDataSource(MAX_CONNECTION, cfg.getString(ApplicationConfigConstants.CF_DSM_DB_URL));
+            String activityDateTime = currentInsertingActivity.getSQLDateTimeString();
+            logger.info(currentInsertingActivity.getDateTimeString());
+            logger.info(currentInsertingActivity.getSQLDateTimeString());
```
SimoneMaiwald: can we make one logger.info

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
+            try (Connection conn = dataSource.getConnection()) {
+                try (PreparedStatement stmt = conn.prepareStatement(INSERT_NEW_ACTIVITIES)) {
+                    stmt.setString(1, kit.getUpsPackage().getUpsPackageId());
+                    stmt.setString(2, kit.getDsmKitRequestId());
+                    stmt.setString(3, currentInsertingActivity.getLocation().getString());
+                    stmt.setString(4, currentInsertingActivity.getStatus().getType());
+                    stmt.setString(5, currentInsertingActivity.getStatus().getDescription());
+                    stmt.setString(6, currentInsertingActivity.getStatus().getCode());
```
SimoneMaiwald: I worry about potential NPEs here

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
+            this.logger.info("Tracking id " + trackingId + " is already delivered, not going to check UPS anymore");
+            updateDeliveryInformation(kit.getUpsPackage(), kit, cfg);
+            return;
         }
-        if (response != null && response.getErrors() == null) {
-            updateStatus(trackingId, type, response, isReturn, kit, ddpInstance);
+        logger.info("Checking UPS status for " + trackingId + " for kit w/ external order number " + kit.getExternalOrderNumber());
+        try {
+            UPSTrackingResponse response = lookupTrackingInfo(trackingId);
+            logger.info("UPS response for " + trackingId + " is " + response);//todo remove after tests
```
SimoneMaiwald: maybe good time to remove that now

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
...
+                "                    q.DSM_KIT_REQUEST_ID = request.dsm_kit_request_id " +
+                "                    and q.event_triggered = true" +
+                "                    )" +
+                "        and (eve.ddp_instance_id = request.ddp_instance_id and eve.kit_type_id = request.kit_type_id) and eve.event_type = ? " +
+                "         and realm.ddp_instance_id = ?" +
+                "          and kit.dsm_kit_request_id = ?";
+        logger.info("Inserting new activities for kit with package id " + kit.getUpsPackage().getUpsPackageId());
+        //        DateFormat dateFormat = new SimpleDateFormat("yyyy-mm-dd");
+        //        String strDate = dateFormat.format(currentInsertingActivity.getDate());
+        //        System.out.println("Converted String: " + strDate);
```
SimoneMaiwald: remove unused code for readability

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -1,289 +1,450 @@
 package org.broadinstitute.dsm.jobs;
```
SimoneMaiwald, line 1: maybe comment that class a bit more. I worry looking at it that in 2-3 month we won't know anymore what is done where and why

src/main/java/org/broadinstitute/dsm/model/UPSKit.java
```diff
@@ -0,0 +1,70 @@
+package org.broadinstitute.dsm.model;
```
SimoneMaiwald: should be moved to package org.broadinstitute.dsm.model.ups

src/main/java/org/broadinstitute/dsm/model/ups/UPSPackage.java
```diff
@@ -30,7 +39,8 @@ public UPSActivity getEarliestFilterredEvent(EventFilter filter) {
             if (filter.includeEvent(event)) {
                 if (earliestFilteredEvent == null) {
                     earliestFilteredEvent = event;
-                } else if (event.getInstant().isBefore(earliestFilteredEvent.getInstant())) {
+                }
+                else if (event.getInstant().isBefore(earliestFilteredEvent.getInstant())) {
```
SimoneMaiwald: NPE maybe?

src/main/java/org/broadinstitute/dsm/shipping/UPSTracker.java
```diff
@@ -50,13 +50,15 @@ public UPSTrackingResponse lookupTrackingInfo(String trackingId) {
...
-               }
+                error = new RuntimeException("couldn't get response from ups tracking of package " + trackingId, e);
+                logger.error(e.getMessage());
+                throw error;
+
+                //               try {
+                //                   Thread.sleep(5 * 1000);
+                //               } catch (InterruptedException interrupted) {
+                //                   logger.warn("Error while waiting for UPS retry", interrupted);
+                //               }
```
SimoneMaiwald: no wait before retry? then remove please

src/main/resources/liquibase/042-updateDB.xml
```diff
@@ -12,13 +12,73 @@
     </changeSet>
     <changeSet author="simone" id="setDeletedFlag">
         <sql>UPDATE ddp_medical_record as m, (
-                SELECT m.medical_record_id from ddp_medical_record m
-                left join ddp_institution i on (i.institution_id = m.institution_id)
-                left join ddp_participant p on (p.participant_id = i.participant_id)
-                left join ddp_instance realm on (p.ddp_instance_id = realm.ddp_instance_id)
-                where realm.instance_name = 'Prostate' and (p.ddp_participant_id like '1771332079%' OR p.ddp_participant_id like '-763171247%') and type = 'INITIAL_BIOPSY' and i.ddp_institution_id != '1' and i.institution_id &#60; 20000) as m2
-            SET m.deleted = 1 WHERE m2.medical_record_id = m.medical_record_id;</sql>
+            SELECT m.medical_record_id from ddp_medical_record m
```
SimoneMaiwald: @pegahtah why was that changeSet changed

src/test/java/org/broadinstitute/dsm/GBFTest.java
```diff
@@ -1,15 +1,30 @@
 package org.broadinstitute.dsm;
 
+import static org.broadinstitute.dsm.careevolve.Covid19OrderRegistrar.ADDRESS_FIELD;
+import static org.broadinstitute.dsm.careevolve.Covid19OrderRegistrar.FIRST_NAME_FIELD;
+import static org.broadinstitute.dsm.careevolve.Covid19OrderRegistrar.GUID_FIELD;
+import static org.broadinstitute.dsm.careevolve.Covid19OrderRegistrar.LAST_NAME_FIELD;
+import static org.broadinstitute.dsm.careevolve.Covid19OrderRegistrar.PROFILE_FIELD;
+
```
SimoneMaiwald: remove unused imports

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
pegahtah: it's the generated key after the update

src/main/java/org/broadinstitute/dsm/cf/Covid19OrderRegistrarFunction.java
pegahtah: This is not my code, it's the original CF code that I branched off of, and changing it won't change the cf unless wee redeploy that. I can change it if you want though

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
pegahtah: It's in the loop, we don't know the problem kit when it's out of the try

src/main/resources/liquibase/042-updateDB.xml
pegahtah: The change was just splitting one long line into 2 lines, I changed it back

src/main/java/org/broadinstitute/dsm/db/DDPInstance.java
pegahtah: On second thoughts, this code is used a lot, I will make a ticket for this change but I'd rather not to do this big change in this ticket.

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -60,184 +65,78 @@ public Covid19OrderRegistrar(String endpoint,
...
             throw new CareEvolveException("Cannot place order for " + kitLabel + " without a pickup time");
         }
 
-        DDPInstance instance = DDPInstance.getDDPInstanceWithRole("testboston", DBConstants.HAS_KIT_REQUEST_ENDPOINTS);
-
-        Map<String, String> queryConditions = new HashMap<>();
-        queryConditions.put("ES", " AND profile.hruid = '" + participantHruid +"'");
-        List<ParticipantWrapper> participants = ParticipantWrapper.getFilteredList(instance, queryConditions);
+        DDPInstance ddpInstance = DDPInstance.getDDPInstanceWithRole("testboston", DBConstants.HAS_KIT_REQUEST_ENDPOINTS, conn);
+        Map<String, Map<String, Object>> esData = ElasticSearchUtil.getSingleParticipantFromES(ddpInstance.getName(), ddpInstance.getParticipantIndexES(), esClient, participantHruid);
```
pegahtah, line 139: @zyme this is where I am concerned about changes

src/main/java/org/broadinstitute/dsm/cf/Covid19OrderRegistrarFunction.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Map;
+import java.util.TimeZone;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import static org.broadinstitute.dsm.statics.ApplicationConfigConstants.DSM_DB_URL;
+
+public class Covid19OrderRegistrarFunction  implements BackgroundFunction<Covid19OrderRegistrarFunction.OrderPayload> {
+
+    private static final Logger logger = Logger.getLogger(TestBostonKitTrackerDispatcher.class.getName());
```
SimoneMaiwald, line 33: why is that not `Logger. getLogger(Covid19OrderRegistrarFunction...`?

PR #121, opened 2021-03-01
DDP-5440 | upload kit whout guid

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-5440

**Changes have been done:**

- Added utility method which checks shortId presented in text file is hruid or legacy short id, **UserUtil.isHruid**
- Added utility methods to fetch user data from Elastic Search by hruid or legacy short id, **ParticipantWrapper.getParticipantByShortId**
- Added method which checks if participant already has kit or not by guid or legacyAltPid, **new KitUploadRoute().checkAndSetParticipantIdIfKitExists**
- fixed outdated tests, added some more
- some code refactored


src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -13,13 +14,15 @@
...
 import org.broadinstitute.dsm.util.ElasticSearchUtil;
+import org.broadinstitute.dsm.util.UserUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.*;
 import java.util.stream.Collectors;
 
 @Getter
+@Setter
```
SimoneMaiwald: Can we change that into `@Data`

src/main/java/org/broadinstitute/dsm/util/UserUtil.java
```diff
@@ -64,6 +64,11 @@
     private static final String DISCARD_SAMPLE_MENU = "discardSamples";
     private static final String PDF_DOWNLOAD_MENU = "pdfDownload";
 
+    public static boolean isHruid(@NonNull String participantId) {
+        final String hruidCheck = "^P\\w{5}$";
+        return participantId.matches(hruidCheck);
+    }
+
```
SimoneMaiwald: Not sure this should go in here. That class is for user as in study staff… I think we should make a ParticipantUtil

gmakharat: Sounds good, will move it, thanks

SimoneMaiwald (approved):

PR #123, opened 2021-03-03
DDP-5159 | filtering invitation accepted date

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-5159

**Changes have been done:**
Switched from `java.util.Date` to `java.time.LocalDateTime`, since `java.util.Date`'s method `getTime()`, to produce time in milliseconds, uses local timezone(system's timezone) which may produce problems in different timezones, applied `java.time.Instant` to `LocalDateTime` to provide UTC timezone and conversion to milliseconds.

used `java.time.temporal.ChronoField` to create proper date format if provided date is partial or only year.


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -31,6 +31,12 @@
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.text.ParseException;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
```
SimoneMaiwald: please remove unused imports

SimoneMaiwald (approved):

PR #126, opened 2021-03-08
DDP-5671 added script to allow bulk cancellation of orders in GBF

zyme:
From time to time, we need to cancel orders with GBF in an adhoc manner.  Rather than fiddling with tests and running the risk of uncommenting `@Ignore`, I thought a CLI would be nice, so I write this.


src/main/java/org/broadinstitute/dsm/model/gbf/CancelRequest.java
```diff
@@ -0,0 +1,13 @@
+package org.broadinstitute.dsm.model.gbf;
+
+import com.google.gson.annotations.SerializedName;
```
zyme, line 3: Added a few model classes to match the GBF APIs

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -96,15 +97,17 @@
 
     private static Executor blindTrustEverythingExecutor;
 
-    public GBFRequestUtil() {
+    static {
+        initTrust();
```
zyme, line 101: Before this change, even if you were calling a static method, you would have had to new up an object, lest you would see an NPE because various static methods assume that `blindTrustEverythingExecutor` has been initialized.  Now it gets initialized automagically.

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -96,15 +97,17 @@
...
+    private static synchronized void initTrust() {
         try {
             if (blindTrustEverythingExecutor == null) {
                 blindTrustEverythingExecutor = Executor.newInstance(SecurityUtil.buildHttpClient());
             }
-        }
-        catch (Exception e) {
+        } catch (Exception e) {
             logger.error("Starting up the blindTrustEverythingExecutor ", e);
-            System.exit(-3);
```
zyme, line 107: This is a bit scary--a constructor that calls `System.exit()` can bring down the entire app!

PR #127, opened 2021-03-12
DDP-5609 | feature add family member

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-5609

**Things have done:**

- added new instance role `add_family_member` using liquibase.
- added if instance has new role `add_family_member` to display settings `DisplaySettingsRoute.java:124`.
- added new route for adding family member data to participant `AddFamilyMemberRoute.java `.
- added two POJO class for family member payload json `AddFamilyMemberPayload.java`, `FamilyMemberDetails`.
- used Optional approach for validations for payload json to provide cleaner code for new add family member route.
- added high level Dao interface to provide general operations on database, created several Dao class implementing that interface, to provide cleaner design and useful for unit tests as well.
- added tests for add family member route, used `@BeforeClass`, `@AfterClass` annotations to setup and clean up environment for tests

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/route/AddFamilyMemberRoute.java
+++ b/src/main/java/org/broadinstitute/dsm/route/AddFamilyMemberRoute.java
@@ -22,7 +22,7 @@ public class AddFamilyMemberRoute extends RequestHandler {
 
     private static final Logger logger = LoggerFactory.getLogger(AddFamilyMemberRoute.class);
 
-    private static final String FIELD_TYPE = "RGP_PARTICIPANTS";
+    static final String FIELD_TYPE = "RGP_PARTICIPANTS";
 
     @Override
     protected Object processRequest(Request request, Response response, String userId) throws Exception {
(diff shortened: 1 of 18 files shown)


src/main/java/org/broadinstitute/dsm/route/AddFamilyMemberRoute.java
```diff
@@ -0,0 +1,70 @@
...
+import org.slf4j.LoggerFactory;
+import spark.Request;
+import spark.Response;
+import spark.utils.StringUtils;
+
+public class AddFamilyMemberRoute extends RequestHandler {
+
+    private static final Logger logger = LoggerFactory.getLogger(AddFamilyMemberRoute.class);
+
+    static final String FIELD_TYPE = "RGP_PARTICIPANTS";
```
SimoneMaiwald: can you move that into FamilyMemberDetails. in case any other study will use the add family feature at one point they would have another form.

gmakharat: We may do it more generic like $realm_PARTICIPANTS?

PR #129, opened 2021-03-16
DDP-5672 longitudinal kit restrictions

zyme:
DDP-5672 is about adding logic that prevents longitudinal kits from being ordered via GBF unless the participant has returned a previous kit within the past 30 days.  This PR does not do the ordering in a CF--it's limited in scope.  What's contained here are:

* Centralized GBF ordering classes [GBFOrderFinder](https://github.com/broadinstitute/ddp-study-manager/pull/129/files#diff-e5df9677c4824d5afc40e15ed748b892c2d885b22fb857b5d1e3f18e7fedcbc6),  [GBFOrderTransmitter](https://github.com/broadinstitute/ddp-study-manager/pull/129/files#diff-f4850ce23860608a2584a3cdb2864f996959b8362c1bc07a8f4ce42c7e76b8c1), and GBFOrderGateKeeper.  GBFOrderGateKeeper is less robust at this point, and will be built out in a subsequent PR.
* Removal of inline ordering via GBF during kit request lookup from DSS

Along the way, a few seemingly unrelated things that made life easier:
* Special ES method to _just_ grab participant address data

A subsequent PR will wire up a CF to call GBFOrderGateKeeper.

Although the target of this PR is `prod_2021_02_24_hotfix`, we should coordinate on an upcoming release branch.

A good place it start is [GBFOrderFinderTest](https://github.com/broadinstitute/ddp-study-manager/pull/129/files#diff-09883bd9903eff89968323c066f773267796b0a07e4df51e3eb457d5f8fd021a).  It shows a few different ordering scenarios.


src/test/java/org/broadinstitute/dsm/TestHelper.java
```diff
@@ -103,9 +106,6 @@ public static void setupDB(boolean setupDDPConfigLookup) {
         cfg = cfg.withValue("portal.port", ConfigValueFactory.fromAnyRef("9999"));
         cfg = cfg.withValue("errorAlert.recipientAddress", ConfigValueFactory.fromAnyRef(""));
 //
-        if (!cfg.getString("portal.environment").startsWith("Local")) {
```
zyme: This might be controversial @SimoneMaiwald ... I found myself fiddling with local configs just to "cheat" and get around this check.

SimoneMaiwald:
latest branch on prod is `prod_2021_02_24_esc` and not `prod_2021_02_24_hotfix`

src/main/java/org/broadinstitute/dsm/db/KitRequestShipping.java
```diff
@@ -82,6 +83,10 @@
     private static final String UPDATE_KIT_AUTHORIZE = "UPDATE ddp_kit kit INNER JOIN(SELECT dsm_kit_request_id, MAX(dsm_kit_id) AS kit_id FROM ddp_kit GROUP BY dsm_kit_request_id) groupedKit " +
             "ON kit.dsm_kit_request_id = groupedKit.dsm_kit_request_id AND kit.dsm_kit_id = groupedKit.kit_id SET authorization = ?, authorization_date = ?, " +
             "denial_reason = ?, authorized_by = ? WHERE kit.dsm_kit_request_id = ?";
+    private static final String MARK_ORDER_AS_TRANSMITTED = "\n"+
+            "update ddp_kit_request set order_transmitted_at = ?\n"+
+            "where\n"+
+            "external_order_number = ?";
```
SimoneMaiwald: can we remove `\n` from queries please

src/main/java/org/broadinstitute/dsm/model/gbf/GBFOrderFinder.java
```diff
@@ -0,0 +1,164 @@
...
+                    // now iterate again to get address
+                    while (rs.previous()) {
+                        String participantGuid = rs.getString(DBConstants.DDP_PARTICIPANT_ID);
+                        String externalOrderId = rs.getString(DBConstants.EXTERNAL_ORDER_NUMBER);
+                        String kitName = rs.getString(DBConstants.EXTERNAL_KIT_NAME);
+                        if (addressForParticipants.containsKey(participantGuid)) {
+                            Address recipientAddress = addressForParticipants.get(participantGuid);
+                            kitsToOrder.add(new SimpleKitOrder(recipientAddress, externalOrderId, kitName, participantGuid));
+                        } else {
+
```
SimoneMaiwald: do we want to log something if it would go into the else?

src/main/java/org/broadinstitute/dsm/model/gbf/GBFOrderGateKeeper.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+
+    public void sendPendingOrders(Connection conn) {
+        for (SimpleKitOrder simpleKitOrder : orderFinder.findKitsToOrder(ddpInstanceName, conn)) {
+            Response orderResponse = transmitter.orderKit(simpleKitOrder.getRecipientAddress(), simpleKitOrder.getExternalKitName(), simpleKitOrder.getExternalKitOrderNumber(), simpleKitOrder.getParticipantGuid());
+
+            if (orderResponse.isSuccess()) {
+                try (PreparedStatement stmt = conn.prepareStatement(UPDATE_TRANSMISSION_DATE)) {
+                    //stmt.setTimestamp(1, Instant.now());
```
SimoneMaiwald: I don't think that should be commented out

src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
```diff
@@ -161,6 +163,13 @@ public Object processRequest(Request request, Response response, String userId)
                         logger.info("placing order with external shipper");
                         ExternalShipper shipper = (ExternalShipper) Class.forName(DSMServer.getClassName(kitRequestSettings.getExternalShipper())).newInstance();
                         shipper.orderKitRequests(orderKits, easyPostUtil, kitRequestSettings, shippingCarrier);
+                        // mark kits as transmitted so that background jobs don't try to double order it
```
SimoneMaiwald: could it be that the background job would pick them up before we set them to transmitted? maybe the method to write them into the ddp_kit_request table should already set them to transmitted?

src/main/java/org/broadinstitute/dsm/util/DDPKitRequest.java
```diff
@@ -173,25 +173,6 @@ public void requestAndWriteKitRequests(List<LatestKitRequest> latestKitRequests)
                                     throw new RuntimeException("Important information for kitRequest is missing");
                                 }
                             }
-                            //TODO PEGAH GET UNORDERED OR NOT FOUND ORDERS AND ADD THEM  TO THE LIST
-                            //                            addOtherUnorderedKitsToList(kitsToOrder);
-
-                            //only order if kit were added to kits to order hash (which should only be if a kit has an external shipper)
-                            if (!kitsToOrder.isEmpty()) {
```
SimoneMaiwald, line 180: could we remove the whole kitsToOrder map

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -813,4 +865,84 @@ private static void rangeQueryBuilder(@NonNull BoolQueryBuilder finalQuery, @Non
...
+
+        public String getLastName() {
+            return lastName;
+        }
+
+        public String getParticipantGuid() {
+            return participantGuid;
+        }
+
+        public String getShippingRecipientName() {
```
SimoneMaiwald: DSM is currently using `mailToName` for labels

src/main/resources/liquibase/DDP-5672-longitudinal-wait.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SimoneMaiwald: can we stay with the same naming convention for the updateDB xmls please

src/test/java/org/broadinstitute/dsm/TestHelper.java
SimoneMaiwald: but it made you double check your config =) so you would have hopefully realized when env was prod instead of dev :)

src/main/resources/liquibase/DDP-5672-longitudinal-wait.xml
```diff
@@ -0,0 +1,49 @@
...
+    </changeSet>
+
+    <changeSet id="andrew" author="DDP-5672-backfill">
+        <!--
+            set approximate order date for tbos gbf orders.  if null, set a bogus date
+            so that new code does not accidentally re-order the kit.
+         -->
+        <sql>
+            update ddp_kit_request_settings set longitudinal_max_wait_days=30 where external_shipper='gbf'
+        </sql>
```
pegahtah: usually because of safe update mode we need to update by the id..., I am not sure if that is the case by liquibase too

src/main/java/org/broadinstitute/dsm/model/gbf/GBFOrderGateKeeper.java
pegahtah: I agree

src/main/java/org/broadinstitute/dsm/model/gbf/GBFOrderGateKeeper.java
```diff
@@ -0,0 +1,49 @@
...
+                    stmt.setString(2, simpleKitOrder.getExternalKitOrderNumber());
+                    int numRowsUpdated = stmt.executeUpdate();
+
+                    // putting 10 things in a kit seems crazy?
+                    if (numRowsUpdated > 10) {
+                        throw new RuntimeException("Updated " + numRowsUpdated + " for order " + simpleKitOrder.getExternalKitOrderNumber() + ".  That seems like a lot?");
```
pegahtah, line 40: I think I didn't understand this, if this is updated based on external order number, isn't 10 a bit too high?

src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
zyme: Yes, this is a possibility.  `GBFOrderFinder` does put a read lock onto ddp_kit_request, but that only protects against multiple instances of the eventual CF.  What we'd need to do here is to put read locks on the orders to be issued during the upload as well.  Would involve doing a re-read on `ddp_kit_request`s in the list here with a read lock, and then double checking that `order_transmission_date` is still `null` in order to place the order.  How's that sound?

src/main/resources/liquibase/DDP-5672-longitudinal-wait.xml
zyme: Yes...if you tell me what that is :)

src/main/resources/liquibase/DDP-5672-longitudinal-wait.xml
zyme: I thought safe mode only kicks in when you have a blank `where` clause?  We'll find out...

src/main/java/org/broadinstitute/dsm/model/gbf/GBFOrderGateKeeper.java
zyme: It's a bit of a guess--what's the maximum number of sub-items in a kit a year from now?  Or should we even check this at all?

pegahtah: I see, so this is taking possible additions of subkits into account. Then yes I think we should check that, because as long as all of them are considered one kit, all of those rows in the table will get the same external order number.

src/main/resources/liquibase/DDP-5672-longitudinal-wait.xml
SimoneMaiwald: 045-updateDB.xml should be that one

src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
SimoneMaiwald: I think easiest way would be to just add a transmitted date when adding kit request to the ddp_kit_request table when uploading

zyme: In that case, there will still be a gap where a separate transaction could grab the orders.  How about if we wrap the whole thing in a separate transaction?  One call to GBF, so either the whole list goes or it fails.  Something like:

```
TransactionWrapper.inTransaction(conn -> {
                    uploadKit(ddpInstance, kitType, kitUploadObjects, kitHasSubKits, kitRequestSettings, easyPostUtil, userIdRequest, kitTypeName,
                            uploadAnyway, invalidAddressList, duplicateKitList, orderKits, specialKitList, upload, kitUploadReason, shippingCarrier, conn);

                    //only order if external shipper name is set for that kit request
                    if (StringUtils.isNotBlank(kitRequestSettings.getExternalShipper())) {
                        try {
                            logger.info("placing order with external shipper");
                            ExternalShipper shipper = (ExternalShipper) Class.forName(DSMServer.getClassName(kitRequestSettings.getExternalShipper())).newInstance();
                            shipper.orderKitRequests(orderKits, easyPostUtil, kitRequestSettings, shippingCarrier);
                            // mark kits as transmitted so that background jobs don't try to double order it
                            for (KitRequest orderKit : orderKits) {
                                KitRequestShipping.markOrderTransmittedAt(conn, orderKit.getExternalOrderStatus(), Instant.now());
                            }
                        }
                        catch (RuntimeException e) {
                            logger.error("Failed to sent kit request order to " + kitRequestSettings.getExternalShipper(), e);
                            response.status(500);
                            return new Result(500, "Failed to sent kit request order to " + kitRequestSettings.getExternalShipper());
                        }
                    }
                });
 ```

File: src/main/java/org/broadinstitute/dsm/util/DDPKitRequest.java (hunk 6 above)
Comment (zyme) [reply src/main/java/org/broadinstitute/dsm/util/DDPKitRequest.java:180]: Done--but beware, this cascaded into even more code being deleted :)

File: src/test/java/org/broadinstitute/dsm/TestHelper.java (hunk 1 above)
Comment (zyme) [reply src/test/java/org/broadinstitute/dsm/TestHelper.java]: I shall revert!

File: src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java [hunk 11]
```diff
@@ -194,56 +198,54 @@ public Object processRequest(Request request, Response response, String userId)
     private void uploadKit(@NonNull DDPInstance ddpInstance, @NonNull KitType kitType, List<KitRequest> kitUploadObjects, boolean kitHasSubKits,
                            @NonNull KitRequestSettings kitRequestSettings, @NonNull EasyPostUtil easyPostUtil, @NonNull String userIdRequest,
                            @NonNull String kitTypeName, boolean uploadAnyway, Map<String, KitRequest> invalidAddressList,
-                           List<KitRequest> duplicateKitList, ArrayList<KitRequest> orderKits, List<KitRequest> specialKitList, Value behavior, String uploadReason, String carrier) {
```
zyme, line 197: This should all be the same, just minus the transaction wrapper.

src/main/java/org/broadinstitute/dsm/util/DDPKitRequest.java
```diff
@@ -100,24 +93,10 @@ public void requestAndWriteKitRequests(List<LatestKitRequest> latestKitRequests)
                                                             String collaboratorParticipantId = KitRequestShipping.getCollaboratorParticipantId(latestKit.getBaseURL(), latestKit.getInstanceID(), latestKit.isMigrated(),
                                                                     latestKit.getCollaboratorIdPrefix(), (String) profile.get("guid"), (String) profile.get("hruid"), kitRequestSettings.getCollaboratorParticipantLengthOverwrite());
 
-                                                            if (kitHasSubKits) {
```
zyme, line 103: A few cases of changing the `if (kitHasSubKits)... else {}` to preserve the `else` logic.

src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
```diff
@@ -178,12 +188,6 @@ public Object processRequest(Request request, Response response, String userId)
             catch (UploadLineException e) {
                 return e.getMessage();
             }
-            catch (FileWrongSeparator e) {
```
zyme, line 181: Flattened this to a single `catch` since it's all subclasses.

zyme:
@SimoneMaiwald @pegahtah ready for re-review (hopefully I got everything).

src/main/java/org/broadinstitute/dsm/model/gbf/GBFOrderFinder.java
```diff
@@ -0,0 +1,204 @@
...
+        } catch (MalformedURLException e) {
+            throw new RuntimeException("Could not initialize es client",e);
+        }
+        GBFOrderFinder orderFinder = new GBFOrderFinder(30, 10000, esClient, "participants_structured.testboston.testboston");
+
+
+        TransactionWrapper.inTransaction(conn -> {
+            Collection<SimpleKitOrder> kits = orderFinder.findKitsToOrder("testboston", conn);
+            for (SimpleKitOrder kit : kits) {
+                System.out.println(kit.getExternalKitOrderNumber());
```
SimoneMaiwald, line 197: do we really want system.out?

src/main/java/org/broadinstitute/dsm/util/DDPKitRequest.java
```diff
@@ -132,28 +111,14 @@ public void requestAndWriteKitRequests(List<LatestKitRequest> latestKitRequests)
...
-                                                    // if the kit type has sub kits > like for promise
+                                                    // if the kit type has sub kits (primarily gbf)
                                                     String collaboratorParticipantId = KitRequestShipping.getCollaboratorParticipantId(latestKit.getBaseURL(), latestKit.getInstanceID(), latestKit.isMigrated(),
                                                             latestKit.getCollaboratorIdPrefix(), participant.getParticipantId(), participant.getShortId(), kitRequestSettings.getCollaboratorParticipantLengthOverwrite());
-                                                    //only testboston for now which is not gen2 so it won't matter
-                                                    if (kitHasSubKits) {
-                                                        List<KitSubKits> subKits = kitRequestSettings.getSubKits();
-                                                        String externalOrderNumber = addSubKits(subKits, kitDetail, collaboratorParticipantId, kitRequestSettings, latestKit.getInstanceID(), null);
```
SimoneMaiwald, line 141: I think you removed too much, I don't see subkits getting written into db anymore

src/main/java/org/broadinstitute/dsm/model/gbf/GBFOrderFinder.java
zyme: Nope!

PR #131, opened 2021-03-17
DDP-5615 | add new family member and copy proband info if checked checkbox

gmakharat:
Tickets:
https://broadinstitute.atlassian.net/browse/DDP-5609
https://broadinstitute.atlassian.net/browse/DDP-5615

**Things have done:**

- added new instance role `add_family_member` using liquibase.
- added if instance has new role `add_family_member` to display settings `DisplaySettingsRoute.java:124`.
- added new route for adding family member data to participant `AddFamilyMemberRoute.java `.
- added two POJO class for family member payload json `AddFamilyMemberPayload.java`, `FamilyMemberDetails`.
- used Optional approach for validations for payload json to provide cleaner code for new add family member route.
- added high level Dao interface to provide general operations on database, created several Dao class implementing that interface, to provide cleaner design and useful for unit tests as well.
- added new route `GetParticipantDataRoute` to get all participant data by guid.
- Introduced dependency inversion approach on `NewParticipantData` (maybe weird prefix but `ParticipantData` already exists and once we finally move to Dao/Dto, Entity/UseCases approach(https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html), we will be able to delete old `ParticipantData` and rename `NewParticipantData`)
- added tests for add family member route, used `@BeforeClass`, `@AfterClass` annotations to setup and clean up environment for tests 

**Maybe it is a little bit mix and duplication of old and new classes but i think this will be good starting point to introduce more flexible approaches like Clean Architecture(https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html) for example.**


src/main/java/org/broadinstitute/dsm/model/ddp/FamilyMemberDetails.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public Map<String, String> toMap() {
+        Map<String, String> familyMemberDetailMap = new HashMap<>();
+        List<Field> declaredFields = Arrays.stream(FamilyMemberDetails.class.getDeclaredFields())
+                .filter(f -> !Modifier.isStatic(f.getModifiers()))
+                .collect(Collectors.toList());
+        for (Field f: declaredFields) {
+            try {
+                familyMemberDetailMap.put(f.getAnnotation(SerializedName.class).value(), (String) f.get(this));
```
gmakharat: Used reflection here to dynamically get instance fields' SerializedNames as Map's key.

src/main/java/org/broadinstitute/dsm/db/dao/participant/data/ParticipantDataDao.java
```diff
@@ -0,0 +1,145 @@
...
+        List<ParticipantDataDto> participantDataDtoList = new ArrayList<>();
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult execResult = new SimpleResult();
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_PARTICIPANT_DATA_BY_PARTICIPANT_ID)) {
+                stmt.setString(1, participantId);
+                try(ResultSet rs = stmt.executeQuery()) {
+                    while (rs.next()) {
+                        participantDataDtoList.add(
+                                new ParticipantDataDto(
+                                        rs.getInt(1),
```
SimoneMaiwald: I am not sure I like not knowing what column we are adding to the object…

src/main/java/org/broadinstitute/dsm/db/dao/participant/data/ParticipantDataDao.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.List;
+import java.util.Optional;
+
+import org.broadinstitute.ddp.db.SimpleResult;
+import org.broadinstitute.dsm.db.dao.Dao;
+import org.broadinstitute.dsm.db.dto.participant.data.ParticipantDataDto;
+
+public class ParticipantDataDao implements Dao<ParticipantDataDto> {
+
+    private static final String SQL_PARTICIPANT_DATA_BY_PARTICIPANT_ID = "SELECT * FROM ddp_participant_data WHERE ddp_participant_id = ?";
```
SimoneMaiwald: I am not sure I like selecting `*` instead of specific columns…

src/main/java/org/broadinstitute/dsm/db/dao/participant/data/ParticipantDataDao.java
gmakharat: good point, will change them.

src/main/java/org/broadinstitute/dsm/db/dao/participant/data/ParticipantDataDao.java
gmakharat: do you mean to just list all the columns instead of `*`?

SimoneMaiwald: yes please

src/main/java/org/broadinstitute/dsm/util/ParticipantUtil.java
```diff
@@ -4,6 +4,7 @@
 
 public class ParticipantUtil {
 
+    public static final String PARTICIPANT_ID = "ddpParticipantId";
```
SimoneMaiwald: change it to DDP_PARTICIPANT_ID please in case we want to add dsm internal participantId later as well

src/main/resources/master-changelog.xml
```diff
@@ -47,4 +47,5 @@
     <include file="liquibase/042-updateDB.xml" relativeToChangelogFile="true"/>
     <include file="liquibase/043-updateDB.xml" relativeToChangelogFile="true"/>
     <include file="liquibase/044-updateDB.xml" relativeToChangelogFile="true"/>
+    <include file="liquibase/045-updateDB.xml" relativeToChangelogFile="true"/>
```
SimoneMaiwald: can we change that to number 46? 45 is already taken =)

SimoneMaiwald (approved):

PR #130, opened 2021-03-17
DDP-5672 Longitudinal Kit Restrictions

zyme:
Updates from the [original PR](https://github.com/broadinstitute/ddp-study-manager/pull/129) to ensure proper target branch (along with other PR feedback).

PR #133, opened 2021-03-19
DDP-5672 Merging back to develop

zyme:
For DDP-5672, merging back to `develop` was very complicated, so I basically copy-pasted the changes to a new branch off `develop`.  Original PR is [here](https://github.com/broadinstitute/ddp-study-manager/pull/130) for reference.


src/main/java/org/broadinstitute/dsm/util/DDPKitRequest.java
```diff
@@ -109,7 +94,26 @@ public void requestAndWriteKitRequests(List<LatestKitRequest> latestKitRequests)
                                                 }
                                             }
                                             else {
-                                                throw new RuntimeException("No participant_index was setup for  " + latestKit.getInstanceName());
+                                                //kit requests from gen2 can be removed after all studies are migrated
```
SimoneMaiwald: this section can be removed, because we do not have any gen2 studies with automated kits anymore

zyme: Done!

SimoneMaiwald (approved):

PR #134, opened 2021-03-22
DDP-5712: Add readonly flag

Irakli-Khaladze:
Added readonly to field_settings in the database.
Added readonly to Fieldsettings class and it's functions


src/main/java/org/broadinstitute/dsm/statics/DBConstants.java
```diff
@@ -276,6 +277,7 @@
     public static final String PDF_DOWNLOAD_RELEASE = "pdf_download_release";
     public static final String PARTICIPANT_STATUS_ENDPOINT = "participant_status_endpoint";
     public static final String ADD_FAMILY_MEMBER = "add_family_member";
+    public static final String SHOW_TAB_FIELDS = "show_tab_fields";
```
SimoneMaiwald: show_group_fields might be better

SimoneMaiwald (approved):

PR #135, opened 2021-03-25
DDP-5739

SimoneMaiwald:
changing TestBostonUPSTrackingJob to not keep transaction open while calling ups for kit status

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
+++ b/src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
@@ -20,6 +20,7 @@ import org.quartz.JobExecutionException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.time.Instant;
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -170,7 +177,7 @@ private static void logError(String trackingId, UPSError[] errors) {
         logger.error(errorString);
     }
 
-    private static void updateStatus(@NonNull Connection conn, String trackingId, String oldType, UPSTrackingResponse response, boolean isReturn, DdpKit kit, DDPInstance ddpInstance) {
+    private static void updateStatus(String trackingId, String oldType, UPSTrackingResponse response, boolean isReturn, DdpKit kit, DDPInstance ddpInstance) {
```
zyme, line 181: Can you re-introduce the `Connection` here or on L206?  This code might be re-usable with the connection, and internal to this method it doesn't look like there is any external IO.

SimoneMaiwald: moved it `updateTrackingInfo` so that when we reuse it we don't need any ups objects

zyme (approved):
Looks good!  This should really help with db lock contention.

PR #143, opened 2021-04-05
DDP-5735: export workflows to ES (RGP/AT)

SimoneMaiwald:
- updating/adding workflow object to workflows at `participants_structured`
- added method to get ISO8601 date string
- fixed json in liquibase update

miscellaneous:
- removed unused imports
- changed Getter/Setter to Data
- removed unused methods


src/main/java/org/broadinstitute/dsm/util/SystemUtil.java
```diff
@@ -35,6 +37,13 @@
     private static final String LINEBREAK = "\r";
     public static final String SEPARATOR = "\t";
 
+    public static String getISO8601DateString() {
+        TimeZone tz = TimeZone.getTimeZone("UTC");
+        DateFormat df = new SimpleDateFormat(ISO_8601_UTC_DATE_STRING);
+        df.setTimeZone(tz);
+        return df.format(new Date());
+    }
+
```
gmakharat, line 43: @SimoneMaiwald I would use everything related to date from `java.time` package, because it was introduced with Java 8 and is much more stable, date related classes from `java.text` are pretty old and they may have problems on new version of Java

src/test/java/org/broadinstitute/dsm/ElasticSearchTest.java
```diff
@@ -684,29 +686,70 @@ public void createTestParticipantsInES() throws Exception {
...
+                newWorkflowMap.put("workflow", "ALIVE_DECEASED");
+                newWorkflowMap.put("status", "ALIVE");
+                newWorkflowMap.put("date", SystemUtil.getISO8601DateString());
+                workflowListES.add(newWorkflowMap);
+                workflowMapES = new HashMap<>();
+                workflowMapES.put("workflows", workflowListES);
+            }
+        }
+        updateES("participants_structured.rgp.rgp", "XLDUNC3BHGWGWERHW781", workflowMapES);
+        Map<String, Object> workflowMapES2 = getObjectByID("participants_structured.rgp.rgp", "XLDUNC3BHGWGWERHW781", "workflows");
```
Irakli-Khaladze, line 725: I guess, you wanted to use assertion here like it was using searchProfileValue before?

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -234,26 +235,52 @@ public static boolean writeWorkflow(@NonNull DDPInstance instance, @NonNull Stri
...
+                    newWorkflowMap.put("date", SystemUtil.getISO8601DateString());
                     List<Map<String, Object>> workflowList = new ArrayList<>();
-                    workflowList.add(workflowMap);
-
-                    Map<String, Object> jsonMap = new HashMap<>();
-                    jsonMap.put(WORKFLOWS, workflowList);
+                    workflowList.add(newWorkflowMap);
+                    workflowMapES = new HashMap<>();
+                    workflowMapES.put("workflows", workflowList);
+                }
```
gmakharat, line 273: @SimoneMaiwald seems like the code from line 239 to 272 is used in tests as well and maybe it would be good idea to extract it as a method just to avoid code duplication?

SimoneMaiwald: I agree =)

gmakharat (approved):

PR #146, opened 2021-04-06
prod_20210406_EllkayHotFix

pegahtah:
Logging the problematic content we receive from ellkay

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
+++ b/src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
@@ -220,7 +220,7 @@ public class Covid19OrderRegistrar {
                             logger.info("Placed CE order {} {} for {}", orderResponse.getHandle(), orderResponse.getHl7Ack(), patientId);
                         }
                         else {
-                            logger.error("Theree was an error while placing order for kitLabel " + kitLabel);
+                            logger.error("There was an error while placing order for kitLabel " + kitLabel);
                         }
                     }
                     catch (IOException e) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -282,11 +288,20 @@ private OrderResponse orderTest(Authentication auth, Message message) throws IOE
...
-            OrderResponse orderResponse = new Gson().fromJson(IOUtils.toString(httpResponse.getEntity().getContent()), OrderResponse.class);
-            return orderResponse;
-        } else {
+            String content = IOUtils.toString(httpResponse.getEntity().getContent());
+            try {
+                OrderResponse orderResponse = new Gson().fromJson(content, OrderResponse.class);
+                return orderResponse;
+            }
+            catch (Exception e) {
+                logger.error("CE response is broken " + content);
```
SimoneMaiwald: How about rephrasing this to : Unexpected CE response

src/main/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrar.java
```diff
@@ -282,11 +288,20 @@ private OrderResponse orderTest(Authentication auth, Message message) throws IOE
         String responseString = EntityUtils.toString(httpResponse.getEntity());
 
         if (httpResponse.getStatusLine().getStatusCode() == HttpStatus.SC_OK) {
-            OrderResponse orderResponse = new Gson().fromJson(IOUtils.toString(httpResponse.getEntity().getContent()), OrderResponse.class);
-            return orderResponse;
-        } else {
+            String content = IOUtils.toString(httpResponse.getEntity().getContent());
```
SimoneMaiwald: maybe do a check here that it is not throwing an error

PR #147, opened 2021-04-07
prod_20210407_NPEHotFix

pegahtah:
NPE fix in case of no confirmation there to receive

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
@@ -313,25 +313,26 @@ public class GBFRequestUtil implements ExternalShipper {
             ShippingConfirmations shippingConfirmations = objectFromXMLString(ShippingConfirmations.class, gbfResponse.getXML());
             if (shippingConfirmations != null) {
                 List<ShippingConfirmation> confirmationList = shippingConfirmations.getShippingConfirmations();
-                if (confirmationList != null) {
+                if (confirmationList != null && !confirmationList.isEmpty()) {
                     Collections.shuffle(confirmationList);
                     logger.info("Number of confirmations received: " + confirmationList.size());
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -311,21 +313,24 @@ public void orderConfirmation(long startDate, long endDate) throws Exception {
...
-                if (confirmationList != null && !confirmationList.isEmpty()) {
-                    for (ShippingConfirmation confirmation : confirmationList) {
-                        try {
-                            processingSingleConfirmation(gbfResponse, confirmation);
-                        } catch (Exception e) {
-                            logger.error("Could not process confirmation for " + confirmation.getOrderNumber(), e);
+                if (confirmationList != null) {
+                    Collections.shuffle(confirmationList);
+                    logger.info("Number of confirmations received: " + confirmationList.size());
+                    if (confirmationList != null && !confirmationList.isEmpty()) {
```
SimoneMaiwald: that should be the check for L316

src/main/java/org/broadinstitute/dsm/util/externalShipper/GBFRequestUtil.java
```diff
@@ -311,23 +313,27 @@ public void orderConfirmation(long startDate, long endDate) throws Exception {
...
+                        catch (Exception e) {
                             logger.error("Could not process confirmation for " + confirmation.getOrderNumber(), e);
                         }
                     }
                     DBUtil.updateBookmark(endDate, DBConstants.GBF_CONFIRMATION); //TODO can be removed because not used anymore
                     logger.info("Finished adding confirmations into db!");
                 }
                 else {
-                    logger.info("No shipping confirmation returned");
+                    logger.info("Zero confirmations received");
```
SimoneMaiwald: Change to: No ShippingConfirmation Elements

PR #150, opened 2021-04-09
DDP-5867: Error on missing ES Index

SimoneMaiwald:
don't throw runtime exception when ES doesn't find the index, only log error


SimoneMaiwald:
This will still log an error in slack but user will see an empty participant list and no red error message

PR #154, opened 2021-04-14
DDP-5735: dsm to ES v3

Irakli-Khaladze:
Main change in this pull request is writeDsmRecord. Was following the approach like workflows. writeDsmRecord is a bit genaralized for both medical records and tissue records. Made small refactoring as well.


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -271,26 +272,79 @@ public static boolean writeWorkflow(@NonNull DDPInstance instance, @NonNull Stri
...
+            UpdateRequest updateRequest = new UpdateRequest()
+                    .index(index)
+                    .type("_doc")
+                    .id(ddpParticipantId)
+                    .doc(objectsMapES)
+                    .docAsUpsert(true);
+
+            UpdateResponse updateResponse = client.update(updateRequest, RequestOptions.DEFAULT);
+
+            logger.info("Updated workflow information for participant " + ddpParticipantId + " in ES for instance " + instance.getName());
```
SimoneMaiwald: can we update the log info to match the method, thank you

SimoneMaiwald (approved):

PR #155, opened 2021-04-15
DDP-5918 | create proband automatically with existing data

gmakharat:
Tickets:
https://broadinstitute.atlassian.net/browse/DDP-5918
https://broadinstitute.atlassian.net/browse/DDP-5921

**Things have done:**

- Added method `setDefaultProbandDataIfNotExists` which checks, whether participant has data or not, if not adds default proband data, checks if data exists but does it contain proband data or not, if not adds default proband data as well, checks if proband data exists but if participant data changed on `DSS` side which should appear in `ES` is the same as already saved data or not, if not it updates `DSM` participant data according to changes in `ES`.
- Added method `extractAndInsertProbandFromESData` which extracts proband default data from `ES` and inserts it to `DSM` database as participant's data.
- Added method `updateProbandDataIfESParticipantUpdated` which checks if participant data was updated on `DSS` side and appeared in `ES`, if so, it updates this participant's data in `DSM` database as well.
- Added classes for constants for related features.


src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -275,6 +289,116 @@ public static String getParticipantLegacyAltPid(Optional<ParticipantWrapper> may
...
+            );
+            new ParticipantDataDao().updateParticipantDataColumn(updatedParticipantDataDTO);
+        }
+    }
+
+    private static void extractAndInsertProbandFromESData(DDPInstance instance, Map<String, Object> esData) {
+        Map<String, Object> profile = (Map<String, Object>) esData.get(ElasticSearchUtil.PROFILE);
+        NewParticipantData newParticipantData;
+        Map<String, String> probandDataMap = extractProbandDefaultDataFromParticipantProfile(esData);
+        newParticipantData = new NewParticipantData(new ParticipantDataDao());
```
SimoneMaiwald: can we make one line out of  L357 and L359

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -93,9 +103,13 @@ public JsonObject getDataAsJson() {
             Map<String, Map<String, Object>> proxyData = getProxyData(instance);
             Map<String, List<ParticipantData>> participantData = ParticipantData.getParticipantData(instance.getName());
 
+            if (DDPInstanceDao.getRole(instance.getName(), DBConstants.ADD_FAMILY_MEMBER)) {
+                participantData = setDefaultProbandDataIfNotExists(participantData, participantESData, instance);
```
SimoneMaiwald, line 108: can we add a comment that this is needed for RGP family member

SimoneMaiwald (approved):

PR #159, opened 2021-04-20
Ddp 5735 sample

Irakli-Khaladze:
Created new class for ES constants, which I'm using during exporting and moved constants to that file. Added dao and dto for ki requests and used them to export sample sent date to elastic search from KitStatusChangeRoute, added a method to export sample fields to ES, made small refactorings


SimoneMaiwald:

PR #160, opened 2021-04-26
DDP 5957: elastic export

Irakli-Khaladze:
This pull request continues pull request 5735

As this pull request was getting too large I decided to create this one, which es responsible to export workflow and medical records data from our database to elastic search. I'll do it for tissues and samples in separate pull request, which will continue this one.
The main changes is that I created subscriber for elastic export, which when the message is published goes to ExportToES class and uses exportObjectsToES method, which itself uses methods for each type of our object for ES to export them. For this changes I also refactored method how writeDsmRecord, where I supply map of keys and values instead of just key and value, so that we don't need to update ES for each field but for all the fields of the object. Other changes are just necessary additions to get all the data from our DB


SimoneMaiwald:

PR #164, opened 2021-04-27
DDP: 5957 elastic export tissue

Irakli-Khaladze:
Added necessary classes to export tissue records to elastic search. Made small changes for writeDsmRecord to work correctly for all the cases


SimoneMaiwald:
I think this one can be closed becuase of the changes from https://github.com/broadinstitute/ddp-study-manager/pull/168

PR #165, opened 2021-04-28
Ddp 5957 elastic export samples

Irakli-Khaladze:
Main change is getting samples from db in KitRequestDao and using it in exportToES


SimoneMaiwald:

PR #171, opened 2021-04-30
prod_20210429b

pegahtah:


SimoneMaiwald (approved):

zyme (approved):
Looks good!

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -436,15 +438,16 @@ private void updateTrackingInfo(PoolingDataSource<PoolableConnection> dataSource
                         if (earliestInTransitTime != null && !kit.getCE_order()) {
                             // if we have the first date of an inbound event, create an order in CE
                             // using the earliest date of inbound event
-                            Authentication careEvolveAuth = null;
-                            if (orderRegistrar == null) {
+
+                            if (orderRegistrar == null || careEvolveAuth == null) {
```
pegahtah, line 442: @zyme here is the change

pegahtah: You can see we added the `careEvolveAuth` to the condition and initialization was moved to L43

zyme, line 442: How about making `orderRegistrar` and `careEvolveAuth` fields like `endpoint` and initializing them in `accept`?  Would be easier to reason through things because we could then know that these two things would always be non-null.

PR #173, opened 2021-05-04
Fix the bug for ordering and inserting

pegahtah:


src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -414,84 +427,85 @@ private void updateTrackingInfo(PoolingDataSource<PoolableConnection> dataSource
...
                         logger.error(r + " is too big for 1 new activity");
                     }
-                    String oldType = null;
-                    if (lastActivity != null && lastActivity.getStatus() != null) {
-                        oldType = lastActivity.getStatus().getType();
+                }
+                catch (Exception ex) {
+                    throw new RuntimeException("Error preparing statement for inserting a new activity  for package id " + kit.getUpsPackage().getUpsPackageId(), ex);
+                }
+            }
```
pegahtah, line 438: loop closes here, still in the `try`

PR #180, opened 2021-05-18
DDP-6154: - fixed sample pages for RGP

SimoneMaiwald:


pegahtah (approved):

SimoneMaiwald:
Let's see when RGP goes pepper, we can close that PR if RGP will go to prod before develop branch does

SimoneMaiwald:
RGP will get migrated in 3 days, so we will not fix the upload for these 3 days

PR #182, opened 2021-05-18
DDP-6141 | Family Id

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-6141

**Things have done**:
1. created liquibase xml file to set initial value in bookmark table for `rgp_family_id` counter.
2. Created bookmark Dao/Dto classes to manipulate bookmark data.
3. Added logic to create `familyId` for proband participants


src/main/resources/liquibase/rgp-family-id-counter.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SimoneMaiwald: can you add the ticket number to the file name please

src/main/java/org/broadinstitute/dsm/db/dao/bookmark/BookmarkDao.java
```diff
@@ -0,0 +1,88 @@
...
+public class BookmarkDao implements Dao<BookmarkDto> {
+
+    private static final String SQL_GET_BOOKMARK_BY_INSTANCE_NAME = "SELECT " +
+            "bookmark_id, " +
+            "value, " +
+            "instance " +
+            "FROM bookmark WHERE instance = ?";
+
+    private static final String SQL_UPDATE_BOOKMARK = "UPDATE " +
+            "bookmark SET value = ?, instance = ? WHERE bookmark_id = ?";
```
SimoneMaiwald, line 23: i don't think we should update the instance here as well.

src/main/resources/master-changelog.xml
```diff
@@ -52,4 +52,5 @@
     <include file="liquibase/048-updateDB.xml" relativeToChangelogFile="true"/>
     <include file="liquibase/049-updateDB.xml" relativeToChangelogFile="true"/>
     <include file="liquibase/050-updateDB.xml" relativeToChangelogFile="true"/>
+    <include file="liquibase/DDP-6141-rgp-family-id-counter.xml" relativeToChangelogFile="true"/>
```
SimoneMaiwald: can we make an `_` between ticket id and short description?

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -375,10 +379,13 @@ private static void updateProbandDataIfESParticipantUpdated(DDPInstance instance
         }
     }
 
-    private static void extractAndInsertProbandFromESData(DDPInstance instance, Map<String, Object> esData) {
+    private static void extractAndInsertProbandFromESData(DDPInstance instance, Map<String, Object> esData,
```
SimoneMaiwald: I wonder if we should move these methods which are used only for study with ADD_FAMILY_MEMBER into a separate study class

src/main/java/org/broadinstitute/dsm/model/rgp/AutomaticProbandDataCreator.java
```diff
@@ -0,0 +1,158 @@
...
+        return ParticipantData.getParticipantData(instance.getName());
+    }
+
+    private static void extractAndInsertProbandFromESData(DDPInstance instance, Map.Entry<String, Map<String, Object>> esData,
+                                                          ParticipantDataDao participantDataDao,
+                                                          BookmarkDao bookmarkDao,
+                                                          Map<String, String> columnsWithDefaultOptions) {
+        String participantId = esData.getKey();
+        NewParticipantData newParticipantData = new NewParticipantData(participantDataDao);
+        Optional<BookmarkDto> maybeBookmark = bookmarkDao.getBookmarkByInstance("rgp_family_id");
```
SimoneMaiwald: can we make that a static string please

src/main/java/org/broadinstitute/dsm/model/rgp/AutomaticProbandDataCreator.java
```diff
@@ -0,0 +1,158 @@
...
+            } else {
+                Optional<ParticipantData> probandData = participantDataList.stream()
+                        .filter(pData -> FamilyMemberConstants.MEMBER_TYPE_SELF.equals(new Gson().fromJson(pData.getData(), Map.class).get(FamilyMemberConstants.MEMBER_TYPE)))
+                        .findFirst();
+                if (probandData.isPresent()) {
+                    updateProbandDataIfESParticipantUpdated(instance, entry, probandData);
+                }
+            }
+        }
+        return ParticipantData.getParticipantData(instance.getName());
```
SimoneMaiwald: can we change it to only get the participantData again, if something was changed in ddp_participant_data? no need to query the db again if nothing was updated

SimoneMaiwald (approved):

PR #184, opened 2021-05-19
Ddp 5735 fix and minor changes

Irakli-Khaladze:
There was a bug caused by new changes in samples, fixed that. Added logs for export, made small refactoring


src/main/java/org/broadinstitute/dsm/export/ExportToES.java
```diff
@@ -4,6 +4,7 @@
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import org.broadinstitute.dsm.DSMServer;
```
SimoneMaiwald: Is that import needed?

Irakli-Khaladze: Deleted

SimoneMaiwald (approved):

PR #185, opened 2021-05-20
DDP-5843: Creating Genome ID

SimoneMaiwald:
for AT new signups need a collaborator id which should be `DDP_ATCP_[HRUID]`
already existing pts which will get added per migration already have that ID so only new participants which will not have any information in ddp_participant_data yet will need to have that id set


src/main/resources/liquibase/DDP-4702_AT-form.xml
```diff
@@ -0,0 +1,481 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SimoneMaiwald, line 1: Not adding that file yet to the master-changelog.xml till we know it is the final version and will stay that way

src/main/java/org/broadinstitute/dsm/model/at/DefaultValues.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DefaultValues {
+
+    private static final Logger logger = LoggerFactory.getLogger(DefaultValues.class);
+
+    private static final String FIELD_TYPE_ID = "AT_GROUP_GENOME_STUDY";
+    private static final String GENOME_STUDY_CPT_ID = "GENOME_STUDY_CPT_ID";
```
SimoneMaiwald, line 22: because AT only has that field which needs to get added and no other default values for the form, don't need to query for the rest of the FieldSettings

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -143,6 +148,15 @@ else if (DBConstants.DDP_KIT_REQUEST_ALIAS.equals(source)) {
                         kitRequests = KitRequestShipping.getKitRequests(instance, filters.get(source));
                         baseList = getCommonEntries(baseList, new ArrayList<>(kitRequests.keySet()));
                     }
+                    else if (DBConstants.DDP_PARTICIPANT_DATA_ALIAS.equals(source)) {
+                        participantData = ParticipantData.getParticipantData(instance.getName(), filters.get(source));
+                        baseList = getCommonEntries(baseList, new ArrayList<>(participantData.keySet()));
+
+                        //if study is AT
+                        if ("atcp".equals(instance.getName())) {
```
gmakharat, line 147: Is it not better to put this if statement on line 230 to avoid code duplication since this if statement is used on line 217 as well

src/main/java/org/broadinstitute/dsm/model/at/DefaultValues.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final String PREFIX = "DDP_ATCP_";
+
+    public static Map<String, List<ParticipantData>> addDefaultValues(Map<String, List<ParticipantData>> participantData,
+                                        Map<String, Map<String, Object>> participantESData, @NonNull DDPInstance instance, String queryAddition) {
+        if (participantESData == null) {
+            logger.warn("Could not update default values, participant ES data is null");
+            return participantData;
+        }
+        boolean addedNewParticipantData = false;
+        for (String ddpParticipantId : participantESData.keySet()) {
```
gmakharat: Usually I prefer to use Map.Entry in for each loop since it returns entry object which has `.getKey()` and `.getValue()` methods which is faster since looping over `.keySet()` and then using `.get()` method of map takes more operations

src/main/java/org/broadinstitute/dsm/model/at/DefaultValues.java
```diff
@@ -0,0 +1,63 @@
...
+            List<ParticipantData> participantDataList = participantData.get(ddpParticipantId);
+            //only needed for new signups - migrated pts already have the values needed
+            if (participantDataList == null) {
+                Map<String, Object> esParticipantData = participantESData.get(ddpParticipantId);
+                Map<String, Object> profile = (Map<String, Object>) esParticipantData.get(ElasticSearchUtil.PROFILE);
+                String hruid = (String) profile.get(ElasticSearchUtil.HRUID);
+                if (StringUtils.isNotBlank(hruid)) {
+                    ParticipantDataDao participantDataDao = new ParticipantDataDao();
```
gmakharat: I would take create of participantDataDao out of for loop, just not to overload memory with create of many objects of ParticipantDataDao since they won't differ from each other

src/main/resources/liquibase/DDP-4702_AT-form.xml
```diff
@@ -0,0 +1,481 @@
...
+                <column name="order_number" value="2"/>
+            </insert>
+            <insert tableName="field_settings">
+                <column name="ddp_instance_id"
+                        valueComputed="(SELECT ddp_instance_id FROM ddp_instance WHERE instance_name='atcp')"></column>
+                <column name="field_type" value="AT_GROUP_ELIGIBILITY"/>
+                <column name="column_name" value="ELIGIBILITY"/>
+                <column name="column_display" value="Eligibility"/>
+                <column name="display_type" value="RADIO"/>
+                <column name="possible_values" value="[{&#34;value&#34;:&#34;2&#34;,&#34;name&#34;:&#34;Pending&#34;},{&#34;value&#34;:&#34;1&#34;,&#34;name&#34;:&#34;Eligible&#34;},{&#34;value&#34;:&#34;0&#34;,&#34;name&#34;:&#34;Not Eligible&#34;}]"/>
```
gmakharat: In the future I would use xml way of escaping quotes which is `&quot;` instead of `&#34;` which would be more easier to find if some typo happened

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
SimoneMaiwald: It will only go into L213 if participantData was not filtered for, so if study staff had a default filter which already filtered for participantData we would not check and add default values

src/main/java/org/broadinstitute/dsm/model/at/DefaultValues.java
SimoneMaiwald: if Map.Entry is faster, I will refactor to use it ☺️ thank you

src/main/java/org/broadinstitute/dsm/model/at/DefaultValues.java
SimoneMaiwald: good point

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
gmakharat: Oh sorry, did not pay attention on `addDefaultValues` method's last argument that it takes query addition.

gmakharat (approved):

PR #192, opened 2021-05-27
DDP-5782: rgp form update

SimoneMaiwald:
- fixed service deletion

- refactored PatchRoute a bit to have static strings and added logic for adding new participants to ddp_participant and ddp_participant_record table (for RGP additional field under participant)

- added RGP form per liqubase but **not yet** to master-changelog.xml


src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -263,6 +272,27 @@ else if (participantDataId != null) {
                         }
                         return new Result(200, new GsonBuilder().serializeNulls().create().toJson(map));
                     }
+                    else if (Patch.DDP_PARTICIPANT_ID.equals(patch.getParent())) {
+                        //new additional value for pt which is not in ddp_participant and ddp_participant_record table yet
+                        DDPInstance ddpInstance = DDPInstance.getDDPInstance(patch.getRealm());
+                        String participantId = inTransaction(conn -> {
+                            String newParticipantId = MedicalRecordUtil.writeParticipantIntoDB(conn, patch.getParentId(), ddpInstance.getDdpInstanceId(), 0, "", patch.getUser());
```
gmakharat: I think it is good time to introduce `DDParticipantDao` and move `writeParticipantIntoDB` method there to avoid DB connection and interaction here.

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -263,6 +272,27 @@ else if (participantDataId != null) {
...
+                            MedicalRecordUtil.writeNewRecordIntoDb(conn, DDPMedicalRecordDataRequest.SQL_INSERT_PARTICIPANT_RECORD, patch.getParentId(), ddpInstance.getDdpInstanceId());
```
gmakharat: the same would be good for `ddp_participant_record`.

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -263,6 +272,27 @@ else if (participantDataId != null) {
...
+                            return newParticipantId;
+                        });
+                        if (StringUtils.isNotBlank(participantId)) {
```
gmakharat: I always prefer to use `Optional`(Optional.ofNullable) as a return type from method which returns single result or empty `List` if it returns collection of records when interaction happens with DB, with `Optional` and empty `List` we can provide more cleaner code and better way to avoid NPEs.

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -263,6 +272,27 @@ else if (participantDataId != null) {
...
+                            DBElement dbElement = patchUtil.getColumnNameMap().get(patch.getNameValue().getName());
+                            if (dbElement != null) {
+                                Patch.patch(participantId, patch.getUser(), patch.getNameValue(), dbElement);
+                            }
+                            else {
+                                throw new RuntimeException("DBElement not found in ColumnNameMap: " + patch.getNameValue().getName());
+                            }
+                            Map<String, String> map = new HashMap<>();
+                            map.put(PARTICIPANT_ID, participantId);
+                            return new Result(200, new GsonBuilder().serializeNulls().create().toJson(map));
```
gmakharat: To be honest, I really don't understand using `Result` instead of directly return result and let `SparkJava` to automatically transform it into json and send it back, if main idea is to set response code we could done it using `response.status()` method for setting specific codes.

SimoneMaiwald: inherited logic from lddp, I agree it is time to change that!

src/main/java/org/broadinstitute/dsm/model/at/DefaultValues.java
```diff
@@ -37,6 +37,7 @@
             if (participantDataList == null) {
                 Map<String, Object> esParticipantData = entry.getValue();
                 Map<String, Object> profile = (Map<String, Object>) esParticipantData.get(ElasticSearchUtil.PROFILE);
+//                TODO only for `registration_type` = `self` && `dependent`
```
SimoneMaiwald: @gmakharat can you add that check in for me.
DSM should only create the genomicId if pt has in ddp_participant_data at field_type_id: `AT_GROUP_MISCELLANEOUS`
`REGISTRATION_TYPE` = `Self` OR `Dependent`

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
SimoneMaiwald: @gmakharat can you please make that change, thank you

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
SimoneMaiwald: @gmakharat same for this :)

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
SimoneMaiwald: feel free to change that

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -263,6 +275,18 @@ else if (participantDataId != null) {
...
+                    else if (Patch.DDP_PARTICIPANT_ID.equals(patch.getParent())) {
+                        //new additional value for pt which is not in ddp_participant and ddp_participant_record table yet
+                        DDPInstance ddpInstance = DDPInstance.getDDPInstance(patch.getRealm());
+                        int participantId = insertDdpParticipant(patch, ddpInstance);
+                        insertDdpParticipantRecord(participantId);
+                        if (participantId > 0) {
+                            DBElement dbElement = patchUtil.getColumnNameMap().get(patch.getNameValue().getName());
+                            if (dbElement == null) throw new RuntimeException("DBElement not found in ColumnNameMap: " + patch.getNameValue().getName());
+                            Patch.patch(String.valueOf(participantId), patch.getUser(), patch.getNameValue(), dbElement);
+                            return new Result(200, new GsonBuilder().serializeNulls().create().toJson(Map.of(PARTICIPANT_ID, String.valueOf(participantId))));
```
gmakharat, line 287: because of lack of time decided to keep `Result` since `PatchRoute` is used widely on frontend side and frontend code uses `result.code`, `result.body` and changing it here would break lots of things on frontend

PR #194, opened 2021-06-04
Ddp 5714 searchable dynamic form fields

Irakli-Khaladze:
Main changes are in FilterRoute and FilterTest. Other stuff is some refactorings.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
+++ b/src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
@@ -316,6 +316,7 @@ public class FilterRoute extends RequestHandler {
                                 .equals(option)) {
                             addParticipantDataCondition(newCondition, first, participantData);
                             participantAdded = true;
+                            first = false;
                             break;
                         }
                     }
--- a/src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
+++ b/src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
@@ -1,11 +1,14 @@
 package org.broadinstitute.dsm.route;
 
 import com.google.gson.Gson;
+import com.google.gson.JsonObject;
 import lombok.NonNull;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.ddp.handlers.util.Result;
 import org.broadinstitute.dsm.db.DDPInstance;
 import org.broadinstitute.dsm.db.ViewFilter;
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
```diff
@@ -238,22 +244,30 @@ else if (savedFilters != null) {
 
     public static List<?> filterParticipantList(Filter[] filters, Map<String, DBElement> columnNameMap, @NonNull DDPInstance instance) {
         Map<String, String> queryConditions = new HashMap<>();
+        List<ParticipantDataDto> allParticipantData = participantDataDao
+                .getParticipantDataByInstanceid(Integer.parseInt(instance.getDdpInstanceId()));
```
SimoneMaiwald: I wonder if we really should have that here. If there is not filter for PARTICIPANT_DATA.equals(filter.getParticipantColumn().tableAlias) then we wouldn't need the list of participantData…

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -91,7 +91,7 @@ public Object processRequest(Request request, Response response, String userId)
                                 if (patch.getActions() != null) {
                                     for (Value action : patch.getActions()) {
                                         if (ESObjectConstants.ELASTIC_EXPORT_WORKFLOWS.equals(action.getType())) {
-                                            writeESWorkflow(patch, nameValue, action);
+//                                            writeESWorkflow(patch, nameValue, action);
```
SimoneMaiwald: why is that commented out?

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -255,7 +255,7 @@ else if (participantDataId != null) {
...
```
SimoneMaiwald: same here

src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
Irakli-Khaladze: Yes, at first I had it inside that if block you mentioned, but in this case we would need to do it for each filter with participantData and that's why moved it there. But I guess there won't be too many filters so maybe it was better to leave it inside the if block not to be called in every case. So do you think I should move it into the if block?

SimoneMaiwald (approved):

PR #196, opened 2021-06-08
DDP-6280 | Update workflow statuses according to the message from DSS…

Irakli-Khaladze:
Created a pub sub subscription, which subscribes message from dss and update workflows for probands in our DB and exports to elastic search. Also created a test to check that update works correctly

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
@@ -443,7 +443,7 @@ public class ElasticSearchUtil {
                     .doc(objectsMapES)
                     .docAsUpsert(true);
 
-//            UpdateResponse updateResponse = client.update(updateRequest, RequestOptions.DEFAULT);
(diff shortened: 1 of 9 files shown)


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -443,7 +443,7 @@ public static void updateRequest(@NonNull String ddpParticipantId, String index,
                     .doc(objectsMapES)
                     .docAsUpsert(true);
 
-            UpdateResponse updateResponse = client.update(updateRequest, RequestOptions.DEFAULT);
+//            UpdateResponse updateResponse = client.update(updateRequest, RequestOptions.DEFAULT);
```
SimoneMaiwald: is the update not needed?

Irakli-Khaladze: Oh, I had it commented out while testing and forgot to bring it back, I'll change it

SimoneMaiwald (approved):

PR #197, opened 2021-06-09
DDP-6280 | generalized workflow subscription

Irakli-Khaladze:
Generalized subscription and made UPDATE_CUSTOM_WORKFLOW as one case of this subscription


src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -379,13 +376,7 @@ private void setupPubSub(@NonNull Config cfg, NotificationUtil notificationUtil)
...
 
         try {
-            ElasticExportSubscription.subscribeElasticExport(projectId, elasticExportSubscriptionId);
-        } catch (Exception e) {
-            e.printStackTrace();
-        }
-
-        try {
-            UpdateWorkflowStatusSubscription.subscribeUpdateWorkflow(projectId, updateWorkflowSubscriptionId);
+            DSMtasksSubscription.subscribeDSMtasks(projectId, DSMtasksSubscriptionId);
```
gmakharat, line 379: Too verbose to have whole pubsub receiver/subscriber setting up from line 330 to 369 in `DSMServer`, I would separate it in different class or in already pubsub existing class, separating database connection/operations on line 336 in `DAO` class would be better as well.

src/main/java/org/broadinstitute/dsm/export/ExportToES.java
```diff
@@ -41,7 +40,21 @@
...
+        public String getStudy() {
+            return study;
+        }
+    }
+
+    public static void exportObjectsToES(String data) {
+        ExportPayload payload = gson.fromJson(data, ExportPayload.class);
         int instanceId = ddpInstanceDao.getDDPInstanceIdByGuid(payload.getStudy());
 
         logger.info("Started exporting workflows for instance with id " + instanceId);
```
gmakharat: Putting logger messages into related method should be more useful than wrapping method in loggers.

src/main/java/org/broadinstitute/dsm/pubsub/DSMtasksSubscription.java
```diff
@@ -5,31 +5,46 @@
...
+
+                    switch (taskType) {
+                        case UPDATE_CUSTOM_WORKFLOW:
+                            WorkflowStatusUpdate.updateCustomWorkflow(attributesMap, data);
+                            break;
+                        case ELASTIC_EXPORT:
+                            ExportToES.exportObjectsToES(data);
+                            break;
+                        default:
+                            logger.warn("Wrong task type for a message from pubsub");
```
gmakharat, line 46: It's always good practice to put break statement even if it is default to avoid possible bugs like if someone put another case below the default, since default does not have break, default content and cases' contents below default will be processed, until it finds break or end of the switch statement.

src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
```diff
@@ -0,0 +1,76 @@
...
+        participantDatas.forEach(participantDataDto -> {
+            updateProbandStatusInDB(workflow, status, participantDataDto, studyGuid);
+        });
+
+        ElasticSearchUtil.writeWorkflow(instance, ddpParticipantId, workflow, status);
+    }
+
+    public static void updateProbandStatusInDB(String workflow, String status, ParticipantDataDto participantDataDto, String studyGuid) {
+        String oldData = participantDataDto.getData();
+        if (oldData != null) {
```
gmakharat: I prefer to use guard conditions to avoid if chaining for better readability.

pegahtah (approved):

PR #199, opened 2021-06-10
Kit upload checks and errors

pegahtah:
Changes so the kit upload is not as strict and errors are more informative


src/test/java/org/broadinstitute/dsm/route/KitUploadRouteTest.java
```diff
@@ -85,7 +81,7 @@ public void fileUserNameDoesNotMatch() {
 
         ParticipantWrapper testParticipant = participantFactory("Mickey", "Mouse", "");
 
-        Assert.assertFalse(route.isKitUploadNameMatchesToEsName(participantFirstNameFromDoc, participantLastNameFromDoc, Optional.of(testParticipant)));
+        Assert.assertNotNull(route.isKitUploadNameMatchesToEsName(participantFirstNameFromDoc, participantLastNameFromDoc, Optional.of(testParticipant)));
```
Irakli-Khaladze: Can it be null? I'm asking because message in isKitUploadNameMatchesToEsName is initialized as empty string

src/test/java/org/broadinstitute/dsm/route/KitUploadRouteTest.java
```diff
@@ -101,7 +97,7 @@ public void fileUserDoesNotBelongToStudy() {
         String participantFirstNameFromDoc = participantDataAsMap.get("firstName");
         String participantLastNameFromDoc = participantDataAsMap.get("lastName");
 
-        Assert.assertFalse(route.isKitUploadNameMatchesToEsName(participantFirstNameFromDoc, participantLastNameFromDoc, Optional.empty()));
+        Assert.assertNotNull(route.isKitUploadNameMatchesToEsName(participantFirstNameFromDoc, participantLastNameFromDoc, Optional.empty()));
```
Irakli-Khaladze: Same question here

src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
```diff
@@ -479,29 +486,39 @@ private int getLastNonEmptyRowIndex(String[] rows) {
...
 
         Optional<ParticipantWrapper> maybeParticipant =
                 ParticipantWrapper.getParticipantByShortId(ddpInstanceByRealm, participantIdFromDoc);
 
         return isKitUploadNameMatchesToEsName(participantFirstNameFromDoc, participantLastNameFromDoc, maybeParticipant);
     }
 
-    boolean isKitUploadNameMatchesToEsName(String participantFirstNameFromDoc, String participantLastNameFromDoc,
-                              Optional<ParticipantWrapper> maybeParticipant) {
+    String isKitUploadNameMatchesToEsName(String participantFirstNameFromDoc, String participantLastNameFromDoc,
```
gmakharat: I would change name as well since the method does not return `boolean` anymore.

src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
```diff
@@ -479,29 +486,39 @@ private int getLastNonEmptyRowIndex(String[] rows) {
...
+        if (!participantFirstNameFromDoc.equals(participantProfile.get("firstName"))) {
+            message += "First names in kit upload don't match the participant";
+            logger.error(message);
+        }
+
+        if (!participantLastNameFromDoc.equals(participantProfile.get("lastName"))) {
+            message += " Last names in kit Upload don't match the participant";
+            logger.error(message);
+        }
+        return message;
```
gmakharat, line 540: Instead of returning message maybe throwing an exceptions directly from if statements would make code more cleaner here and on line 433 as well, on line 433 would have only one `try/catch` block to handle exceptions.

src/test/java/org/broadinstitute/dsm/route/KitUploadRouteTest.java
pegahtah: Good catch! Thanks!

src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
pegahtah: because that way the message we send to the frontend will show "RuntimeException".

src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
```diff
@@ -479,29 +497,48 @@ private int getLastNonEmptyRowIndex(String[] rows) {
...
+            participantProfile.put("lastName", participantProfileFromEs.get("lastName").trim().toLowerCase());
         });
-        return participantFirstNameFromDoc.equals(participantProfile.get("firstName"))
-                && participantLastNameFromDoc.equals(participantProfile.get("lastName"));
+        String message = "";
+
+
+        if (!participantFirstNameFromDoc.equals(participantProfile.get("firstName"))) {
+            message += "First names ";
+            logger.error(message);
```
gmakharat: logger here will log only "First names ".

src/main/java/org/broadinstitute/dsm/route/KitUploadRoute.java
```diff
@@ -479,29 +497,48 @@ private int getLastNonEmptyRowIndex(String[] rows) {
...
+
+        if (!participantLastNameFromDoc.equals(participantProfile.get("lastName"))) {
+            if (StringUtils.isNotBlank(message)) {
+                message += "and ";
+            }
+            message += " Last names ";
+            logger.error(message);
+        }
+        if (StringUtils.isNotBlank(message)) {
+            message += "in kit Upload don't match the participant";
```
gmakharat, line 537: maybe it is better to have logger only here to have fully understandable log message.

PR #200, opened 2021-06-15
DDP-4952: making pdf bundle download 

pegahtah:


Irakli-Khaladze (approved):

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -44,100 +46,69 @@
...
     @Override
     public Object processRequest(Request request, Response response, String userId) throws Exception {
         logger.info(request.url());
-        if (request.url().contains(RoutePath.DOWNLOAD_PDF)) {
-            String realm = null;
-            QueryParamsMap queryParams = request.queryMap();
-            if (queryParams.value(RoutePath.REALM) != null) {
-                realm = queryParams.get(RoutePath.REALM).value();
-            }
+        QueryParamsMap queryParams = request.queryMap();
```
gmakharat, line 67: First of all I would create class something like `PDFPayload` which would handle payload from frontend since bunch of information comes in request regarding pdf download.

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -44,100 +46,69 @@
...
-                        if (request.url().endsWith(CONSENT_PDF)) {
-                            pdfType = CONSENT_PDF;
-                        }
-                        else if (request.url().endsWith(RELEASE_PDF)) {
-                            pdfType = RELEASE_PDF;
-                        }
-                        else if (request.url().contains(COVER_PDF)) {
-                            pdfType = COVER_PDF;
-                            medicalRecord = request.params(RequestParameter.MEDICALRECORDID);
+                if (request.url().contains(RoutePath.DOWNLOAD_PDF)) {
```
gmakharat, line 74: this class in general does lots of things which is not a good idea, makes code so hard to read and debug with tons of if chaining, one of the good way would be to introduce abstraction about PDFs and separate each type of pdf creation logic in different classes under this abstraction, which would provide much more cleaner and understandable code

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -44,100 +46,69 @@
...
+                        if (!userId.equals(userIdR)) {
+                            throw new RuntimeException("User id was not equal. User Id in token " + userId + " user Id in request " + userIdR);
                         }
-                        else if (request.url().contains(REQUEST_PDF)) {
-                            pdfType = REQUEST_PDF;
-                            if (queryParams.value("requestId") != null) {
-                                oncHistoryIDs = Arrays.asList(queryParams.get("requestId").values());
+
+                        String ddpParticipantId = (String) jsonObject.get(RequestParameter.DDP_PARTICIPANT_ID);
+                        if (StringUtils.isNotBlank(ddpParticipantId)) {
```
gmakharat, line 85: handling empty/null values can be handled cleaner using Optionals, `PDFPayload` would have Optional type instance fields thus it would Optional benefits such as `ifPresent`, `orElse`, `ifPresentOrElse` and etc. to avoid such a lot of if chaining if we would not have value in payload.
Second better approach is guard conditions:
`if (SrtingUtils.isBlank(ddpParticipantId)) throw new RuntimeException("Error missing ddpParticipantId");`
`//use ddpParticipantId since we are pretty sure that at this point it is not blank`

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -44,100 +46,69 @@
...
+                                configName = (String) jsonObject.get(RequestParameter.CONFIG_NAME);
                             }
-                        }
-                        else if (request.url().endsWith(PDF) && StringUtils.isNotBlank(configName)) {
-                            pdfType = configName;
-                        }
-                        if (StringUtils.isNotBlank(pdfType)) {
-                            getPDFs(response, ddpParticipantId, pdfType, medicalRecord, oncHistoryIDs, realm, requestBody, userIdRequest);
+                            String medicalRecord = null;
+                            if (jsonObject.has("medicalRecordId")) {
```
gmakharat, line 91: All this getting value logic of ifs from 87 to 98 can be wrapped in `PDFPayload` with using Gson's `fromJson`.

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -146,17 +117,18 @@ else if (request.url().endsWith(PDF) && StringUtils.isNotBlank(configName)) {
...
                                 ElasticSearchUtil.BY_GUID + ddpParticipantId);
-                        if (participantESData == null && participantESData.isEmpty()) {
+                        //filter ES with GUID
+                        if (participantESData != null && !participantESData.isEmpty()) {
+                            return returnPDFS(participantESData, ddpParticipantId);
+                        }
+                        else {
+                            //pt was not found with GUID check for legacyAltPid
                             participantESData = ElasticSearchUtil.getFilteredDDPParticipantsFromES(instance, ElasticSearchUtil.BY_LEGACY_ALTPID + ddpParticipantId);
+                            return returnPDFS(participantESData, ddpParticipantId);
```
gmakharat, line 135: getting participant's all PDFs logic would be separated in its own class

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -197,235 +169,226 @@ public static Object returnPDFS(Map<String, Map<String, Object>> participantESDa
...
-                rawResponse.getOutputStream().close();
-            }
-            else {
-                throw new RuntimeException("Got null back for " + ddpParticipantId + " of ddpInstance " + ddpInstance.getName());
             }
         }
+        addDDPParticipantDataToValueMap(ddpInstance, ddpParticipantId, valueMap, true);
+
+        valueMap.put(CoverPDFProcessor.START_DATE_2, StringUtils.isNotBlank(startDate) ? startDate : valueMap.get(CoverPDFProcessor.FIELD_DATE_OF_DIAGNOSIS)); //start date
+        InputStream stream = null;
```
gmakharat, line 267: You could use try-with-resource to work with auto closables (Input/Output streams) which will automatically close resources.
try( InputStream stream = processor.generateStream(valueMap) ) {
      //stuff do to here
} catch (Exception e) {
     //handling exception
}

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -197,235 +169,226 @@ public static Object returnPDFS(Map<String, Map<String, Object>> participantESDa
...
+    private byte[] tissueRequestPDF(@NonNull DDPInstance ddpInstance, @NonNull String ddpParticipantId, @NonNull List<String> oncHistoryIDs) {
+        logger.info("Generating request pdf for onc history ids {}", StringUtils.join(oncHistoryIDs, ","));
+        RequestPDFProcessor processor = new RequestPDFProcessor(ddpInstance.getName());
+        Map<String, Object> valueMap = new HashMap<>();
+        String today = new SimpleDateFormat("MM/dd/yyyy").format(new Date());
+        valueMap.put(RequestPDFProcessor.FIELD_DATE, today);
+        valueMap.put(RequestPDFProcessor.FIELD_DATE_2, "(" + today + ")");
+        addDDPParticipantDataToValueMap(ddpInstance, ddpParticipantId, valueMap, false);
+
+        InputStream stream = null;
```
gmakharat, line 297: the same about try-with-resource

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -197,235 +169,226 @@ public static Object returnPDFS(Map<String, Map<String, Object>> participantESDa
...
+     *
+     * @throws Exception
+     */
+    public void getPDFs(@NonNull Response response, @NonNull String ddpParticipantId, String pdfType, String medicalRecordId,
+                        List<String> oncHistoryIDs, List<PDF> pdfs, String realm, String requestBody, Integer userId) {
+        DDPInstance ddpInstance = DDPInstance.getDDPInstance(realm);
+        if (ddpInstance != null && StringUtils.isNotBlank(ddpParticipantId)) {
+            byte[] pdfBytes = null;
+            String fileName = "";
+            if (COVER.equals(pdfType)) {
```
gmakharat, line 353: with PDF abstraction you would introduce abstract factory design pattern,  which would help to determine which PDF object needed to be created using `pdfType` for PDF processing.

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -197,235 +169,226 @@ public static Object returnPDFS(Map<String, Map<String, Object>> participantESDa
...
-                    InstanceSettings instanceSettings = InstanceSettings.getInstanceSettings(realm);
-                    if (instanceSettings != null && instanceSettings.getMrCoverPdf() != null && !instanceSettings.getMrCoverPdf().isEmpty()) {
-                        for (Value mrCoverSetting : instanceSettings.getMrCoverPdf()) {
-                            if (keySet.contains(mrCoverSetting.getValue())) {
-                                valueMap.put(mrCoverSetting.getValue(), BooleanUtils.toBoolean((Boolean) jsonObject.get(mrCoverSetting.getValue())));
+    public byte[] getPDFBundle(@NonNull String ddpParticipantId, String medicalRecordId,
+                             List<String> oncHistoryIDs, List<PDF> pdfs, @NonNull String realm, String requestBody) {
+        DDPInstance ddpInstance = DDPInstance.getDDPInstance(realm);
+        if (ddpInstance != null && StringUtils.isNotBlank(ddpParticipantId)) {
+            ByteArrayOutputStream output = new ByteArrayOutputStream();
```
gmakharat, line 184: try to always close resource streams either using try-with-resource or their `close()` method, try-with-resource is always preferable.

PR #201, opened 2021-06-16
Ddp 5730 default participant view list

Irakli-Khaladze:


src/main/resources/liquibase/DDP-5730_default-column-fields.xml
```diff
@@ -0,0 +1,22 @@
...
+        <addColumn tableName="instance_settings">
+            <column name="default_columns" type="VARCHAR(1000)"
+                    afterColumn="study_specific_statuses" />
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="RGPdefaultColumns" author="ikhaladz" >
+        <update tableName="instance_settings">
+            <column name="default_columns" value="[{&quot;value&quot;:&quot;REALM&quot;},{&quot;value&quot;:&quot;SHORT_ID&quot;},{&quot;value&quot;:&quot;ENROLLMENT_STATUS&quot;}]"/>
+            <where>ddp_instance_id = '19'</where>
```
gmakharat: `ddp_instance_id` of RGP will be different for different servers(local, dev, prod) better to use sql statement to get its id by some RGP unique value.

gmakharat (approved):

PR #202, opened 2021-06-17
DDP-6319 | Export default options to ES

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-6319

**Things have done:**

1. Added `getColumnsWithDefaultOptionsFilteredByElasticExportWorkflow` method which filters columns by `ELASTIC_EXPORT.workflows` and added tests for it
2. Refactored code a little bit.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/model/fieldsettings/FieldSettings.java
+++ b/src/main/java/org/broadinstitute/dsm/model/fieldsettings/FieldSettings.java
@@ -3,17 +3,19 @@ package org.broadinstitute.dsm.model.fieldsettings;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import com.google.gson.Gson;
 import lombok.NonNull;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsm.db.dto.fieldsettings.FieldSettingsDto;
+import org.broadinstitute.dsm.model.Value;
+import org.broadinstitute.dsm.statics.ESObjectConstants;
 
 public class FieldSettings {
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/dsm/model/fieldsettings/FieldSettings.java
```diff
@@ -6,12 +6,14 @@
...
 import com.google.gson.Gson;
 import lombok.NonNull;
+import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsm.db.dto.fieldsettings.FieldSettingsDto;
 
 public class FieldSettings {
 
     public static final String KEY_DEFAULT = "default";
     public static final String KEY_VALUE = "value";
+    public static final String ELASTIC_EXPORT_WORKFLOWS = "ELASTIC_EXPORT.workflows";
```
Irakli-Khaladze: You could have used this field from ESObjectConstants

src/main/java/org/broadinstitute/dsm/model/fieldsettings/FieldSettings.java
```diff
@@ -45,4 +57,11 @@ boolean isDefaultOption(String possibleValuesJson) {
          return isDefault;
     }
 
+    boolean isElasticExportWorkflowType(String action) {
+        if (StringUtils.isBlank(action)) return false;
+        List<Map<String, String>> actions = new Gson().fromJson(action, List.class);
```
Irakli-Khaladze: Use Value[] instead of List<Map<String, String>> because actions can have type, type2 and so one and not be key and value pair

Irakli-Khaladze (approved):

PR #203, opened 2021-06-17
Add new participantData in ddp_participant_data and export workflows …

Irakli-Khaladze:
Add new participantData in ddp_participant_data and export workflows only if necessary

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
+++ b/src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
@@ -158,11 +158,11 @@ public class FieldSettingsDao implements Dao<FieldSettingsDto> {
         return fieldSettingsByOptions;
     }
 
-    public Optional<FieldSettingsDto> getFieldSettingByColumnNameAndInstanceId(String columnName, String instanceId) {
+    public Optional<FieldSettingsDto> getFieldSettingByColumnNameAndInstanceId(int instanceId, String columnName) {
         SimpleResult results = inTransaction((conn) -> {
             SimpleResult dbVals = new SimpleResult();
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
```diff
@@ -154,4 +157,46 @@ public int delete(int id) {
...
+            catch (SQLException ex) {
+                execResult.resultException = ex;
+            }
+            return execResult;
+        });
+        if (results.resultException != null) {
+            throw new RuntimeException("Error getting fieldSettings ", results.resultException);
+        }
+        if (fieldSettingsByOptions.size() != 1) {
+            throw new RuntimeException("More than one row for column name " + columnName);
```
gmakharat: Is it possible to have more than 1 record in DB with the same column name?

src/main/java/org/broadinstitute/dsm/pubsub/WorkflowStatusUpdate.java
```diff
@@ -32,11 +39,48 @@ public static void updateCustomWorkflow(Map<String, String> attributesMap, Strin
...
+            }
         });
+    }
 
-        ElasticSearchUtil.writeWorkflow(instance, ddpParticipantId, workflow, status);
+    public static int addNewParticipantDataWithStatus(String workflow, String status, String ddpParticipantId) {
+        Optional<FieldSettingsDto> fieldSetting = fieldSettingsDao.getFieldSettingByColumnName(workflow);
+        JsonObject dataJsonObject = new JsonObject();
+        dataJsonObject.addProperty(workflow, status);
+        AtomicInteger participantDataId = new AtomicInteger();
```
gmakharat: If you wanted to avoid using of AtomicInteger you could use `map` of optional, like:
fieldSetting  
    \.map(setting -> {//dao/dto stuff here})
    \.orElse(0)

gmakharat (approved):

PR #212, opened 2021-06-25
DDP-6473 | Get guids of altpids from ES and update with those guids

Irakli-Khaladze:
When participants have altpids instead of guids, we are getting guid from ES ans updating workflows and family Id-s with it

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/export/ExportToES.java
+++ b/src/main/java/org/broadinstitute/dsm/export/ExportToES.java
@@ -44,8 +44,6 @@ public class ExportToES {
     public static final String FAMILY_ID = "FAMILY_ID";
     public static final String RGP_PARTICIPANTS = "RGP_PARTICIPANTS";
 
-    private Dao dataAccess;
-
     public static class ExportPayload {
         private String index;
         private String study;
@@ -138,7 +136,7 @@ public class ExportToES {
                 continue;
             }
             String ddpParticipantId = participantData.getDdpParticipantId();
-            if (ddpParticipantId.length() > 20) {
+            if (!ParticipantUtil.isGuid(ddpParticipantId)) {
                 ddpParticipantId = getGuidIfWeHaveAltpid(ddpParticipantId, maybeEsParticipantIndex);
             }
             if ("".equals(ddpParticipantId)) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/dsm/export/ExportToES.java
```diff
@@ -40,6 +44,8 @@
     public static final String FAMILY_ID = "FAMILY_ID";
     public static final String RGP_PARTICIPANTS = "RGP_PARTICIPANTS";
 
+    private Dao dataAccess;
```
gmakharat: what's the point of adding this if you don't use?

src/main/java/org/broadinstitute/dsm/export/ExportToES.java
```diff
@@ -124,11 +130,20 @@ private static void exportTissueRecords(int instanceId) {
...
         DDPInstance ddpInstance = DDPInstance.getDDPInstanceById(instanceId);
+        Optional<String> maybeEsParticipantIndex =
+                new DDPInstanceDao().getEsParticipantIndexByInstanceId(instanceId);
         for (ParticipantDataDto participantData: allParticipantData) {
             String data = participantData.getData();
             if (data == null) {
                 continue;
             }
+            String ddpParticipantId = participantData.getDdpParticipantId();
+            if (ddpParticipantId.length() > 20) {
```
gmakharat: there is a method already in `ParticipantUtil` `isGuid` which returns boolean whether id is GUID or not, which will bring more clarification what you are doing here.

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -224,12 +224,40 @@ public static ElasticSearch getParticipantESDataByParticipantId(@NonNull String
         return elasticSearch;
     }
 
+    public static ElasticSearch getParticipantESDataByAltpid(@NonNull String index, @NonNull String altpid) {
```
gmakharat, line 227: Looks like just duplication of `fetchESDataByParticipantId` since `fetchESDataByParticipantId` method determines which to use `altpid` or `guid` as a filter according to `participantId`.

Irakli-Khaladze: This one only uses altpid. fetchESDataByParticipantId wouldn't be relevant in this case. The same staff those methods have is in separate method

gmakharat (approved):

PR #215, opened 2021-06-28
DDP-6484 test boston manual orders

zyme:
DDP-6484 provides a one-off java CLI for placing CareEvolve orders for kits that arrive without UPS data.  See DDP-6484 for more details.

To test this on dev, I used the following set of args: 
`P02133 TBOS-100006T XWPE86O8WIYS2XWO228J "01/01/1970 00:00"`

As for the ["test" class](https://github.com/broadinstitute/ddp-study-manager/pull/215/files#diff-56d817cc09154c43636e68988e842c8a7bbd907e5d31bff46a60d77fc3d596c5), it was a temporary measure for placing orders.  The new CLI is much better, so I've removed the "test".


src/test/java/org/broadinstitute/dsm/careevolve/Covid19OrderRegistrarTest.java
```diff
@@ -1,103 +0,0 @@
...
-        cfg = ConfigFactory.load();
-        TransactionWrapper.init(20, cfg.getString("portal.dbUrl"), cfg, true);
-
-        // todo pull this out to a file, refresh from secret manager
-
-    }
-
-    @Ignore
-    @Test
-    public void testOrderForParticipant() throws Exception {
```
zyme, line 55: The only thing going on here was placing an order, so in the spirit of not doing stuff like this in a "test", it's gone!

src/main/java/org/broadinstitute/dsm/model/birch/RegisterBirchOrderCLI.java
```diff
@@ -0,0 +1,138 @@
...
+            }
+            String participantHruid = args[0];  // user.hruid
+            String kitLabel = args[1]; // ddp_kit.kit_label
+            String externalOrderNumber = args[2];  // ddp_kit_request.external_order_number
+            /** {@link DATE_FORMAT} */
+            String collectionTime = args[3];
+
+            Instant collectionDate = null;
+            try {
+                collectionDate = new SimpleDateFormat(DATE_FORMAT).parse(collectionTime).toInstant();
```
pegahtah: Do we need to change that to Eastern time zone?

zyme: That is a great idea!  It's defaulting to that, but only because of my laptop's timezone...

pegahtah (approved):

src/main/java/org/broadinstitute/dsm/model/birch/RegisterBirchOrderCLI.java
```diff
@@ -0,0 +1,138 @@
...
+            } catch(ParseException e) {
+                throw new RuntimeException("Could not parse collection time '"  + collectionTime + "'.  Please use " + DATE_FORMAT + ".  You may need to quote the whole time.");
+            }
+
+            Config cfg = ConfigFactory.load();
+
+            System.out.println("About to order " + kitLabel + " for participant " + participantHruid + " with collection time " + collectionTime + " and order number " + externalOrderNumber  +".  Ctrl-c to abort...");
+            try {
+                Thread.sleep(5_000);
+            } catch(InterruptedException e) { }
```
yufengwng: One thing I like to do is prompt for confirmation. It does involve more work from the person, but if it's a one-time thing might as well give them more time to think it through. (Or if they're absolutely sure, then they can proceed without having to wait 5 seconds.) It can be a simple "press enter" or can even verify that they said yes. Something like:
```java
            System.out.print("... Continue? [y/N] ");
            Scanner scanner = new Scanner(System.in);
            String input = scanner.nextLine().trim();
            if (!"y".equalsIgnoreCase(input) && !"yes".equalsIgnoreCase(input)) {
                return;
            }
```

src/main/java/org/broadinstitute/dsm/model/birch/RegisterBirchOrderCLI.java
yufengwng: Is this the way to do it?
```java
LocalDateTime.parse(collectionTime, DateTimeFormatter.ofPattern(DATE_FORMAT))
    .atZone(ZoneId.of("America/New_York"))
    .toInstant();
````

**Review by yufengwng — COMMENTED** [review]
Nice CLI! I like that we take in a config file and it reuses a lot of the existing `Covid19OrderRegistrar` functionality (from what I can tell) for placing an order. Just a few suggestions below.

File: src/main/java/org/broadinstitute/dsm/model/birch/RegisterBirchOrderCLI.java [hunk 4]
```diff
@@ -0,0 +1,134 @@
...
+ * CLI utility for manually placing orders in CareEvolve (and birch)
+ * in response to BQMS tickets for kits that have been returned
+ * without UPS shipping info.
+ */
+public class RegisterBirchOrderCLI {
+
+    private static final String USAGE = "args in order are hruid, kitLabel, externalOrderNumber, and collectionTime\n" +
+            "for example java -Dconfig.file=... RegisterBirchOrderCLI P1234 TBOS-123 12345678910111213 06/30/2020 15:34\n" +
+            "Make sure you are on a broad network\n";
+    public static final String DATE_FORMAT = "MM/dd/yyyy HH:mm";
```
zyme, line 43: @yufengwng your suggestion worked once I updated this from `hh` to `HH` (I've been putting in hours as 0-23).

zyme:
Ready for re-review

src/main/java/org/broadinstitute/dsm/model/birch/RegisterBirchOrderCLI.java
yufengwng: Noted! And it's helpful that the "usage" string shows an example of `15:34` hinting at 24-hr format!

yufengwng (approved):
LGTM

pegahtah (approved):

PR #220, opened 2021-07-02
Change logic to check datstat altpid instead of member type

Irakli-Khaladze:


gmakharat (approved):

SimoneMaiwald (approved):
only comment: maybe add a test for clearBeforeUpdate = true

PR #221, opened 2021-07-05
DDP-6502  | add family member export copied proband data to es

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-6502

**Things have done:**

- Added `exportDataToEs` method in `AddFamilyMemberRoute`, if copy proband info is checked it exports copied proband data to ES for family member otherwise it exports default values to ES for family member
- Code refactored to make it more cleaner

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/DDPInstanceDao.java
+++ b/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/DDPInstanceDao.java
@@ -246,7 +246,7 @@ public class DDPInstanceDao implements Dao<DDPInstanceDto> {
                     }
                 }
                 catch (SQLException e) {
-                    throw new RuntimeException("Error getting participant es index with study guid: " + studyGuid, e);
+                    throw new RuntimeException("Error getting collaborator id prefix with study guid: " + studyGuid, e);
                 }
             }
             catch (SQLException ex) {
@@ -256,7 +256,7 @@ public class DDPInstanceDao implements Dao<DDPInstanceDto> {
         });
 
         if (results.resultException != null) {
-            throw new RuntimeException("Couldn't get participant es index with study guid: " + studyGuid, results.resultException);
+            throw new RuntimeException("Couldn't get collaborator id prefix with study guid: " + studyGuid, results.resultException);
         }
         return Optional.ofNullable((String) results.resultValue);
     }
--- a/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/DDPInstanceDao.java
+++ b/src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/DDPInstanceDao.java
@@ -59,6 +59,11 @@ public class DDPInstanceDao implements Dao<DDPInstanceDto> {
             "FROM ddp_instance " +
             "WHERE study_guid = ?";
 
+    private static final String SQL_GET_COLLABORATOR_ID_PREFIX_BY_STUDY_GUID = "SELECT collaborator_id_prefix " +
+            "FROM ddp_instance " +
+            "WHERE study_guid = ?";
+    public static final String COLLABORATOR_ID_PREFIX = "collaborator_id_prefix";
+
     public static boolean getRole(@NonNull String realm, @NonNull String role) {
         SimpleResult results = inTransaction((conn) -> {
             SimpleResult dbVals = new SimpleResult();
@@ -230,5 +235,29 @@ public class DDPInstanceDao implements Dao<DDPInstanceDto> {
         return Optional.ofNullable((String) results.resultValue);
     }
 
+    public Optional<String> getCollaboratorIdPrefixByStudyGuid(String studyGuid) {
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult dbVals = new SimpleResult();
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_COLLABORATOR_ID_PREFIX_BY_STUDY_GUID)) {
+                stmt.setString(1, studyGuid);
+                try (ResultSet rs = stmt.executeQuery()) {
+                    if (rs.next()) {
+                        dbVals.resultValue = rs.getString(COLLABORATOR_ID_PREFIX);
+                    }
+                }
+                catch (SQLException e) {
+                    throw new RuntimeException("Error getting participant es index with study guid: " + studyGuid, e);
+                }
+            }
+            catch (SQLException ex) {
+                dbVals.resultException = ex;
+            }
+            return dbVals;
+        });
 
+        if (results.resultException != null) {
+            throw new RuntimeException("Couldn't get participant es index with study guid: " + studyGuid, results.resultException);
+        }
+        return Optional.ofNullable((String) results.resultValue);
+    }
 }
--- a/src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
+++ b/src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
@@ -101,21 +101,20 @@ public class FieldSettingsDao implements Dao<FieldSettingsDto> {
                 try(ResultSet rs = stmt.executeQuery()) {
                     while (rs.next()) {
                         fieldSettingsByOptions.add(
-                                new FieldSettingsDto(
-                                        rs.getInt(FIELD_SETTINGS_ID),
-                                        rs.getInt(DDP_INSTANCE_ID),
-                                        rs.getString(FIELD_TYPE),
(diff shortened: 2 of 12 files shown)


Irakli-Khaladze (approved):

src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
```diff
@@ -177,21 +177,20 @@ public int delete(int id) {
...
-                                fieldSettingsByColumnNameRs.getString(POSSIBLE_VALUES),
-                                fieldSettingsByColumnNameRs.getString(ACTIONS),
-                                fieldSettingsByColumnNameRs.getBoolean(READONLY),
-                                fieldSettingsByColumnNameRs.getInt(ORDER_NUMBER),
-                                fieldSettingsByColumnNameRs.getBoolean(DELETED),
-                                fieldSettingsByColumnNameRs.getLong(LAST_CHANGED),
-                                fieldSettingsByColumnNameRs.getString(CHANGED_BY)
-                        );
+                        dbVals.resultValue = new FieldSettingsDto.Builder(fieldSettingsByColumnNameRs.getInt(DDP_INSTANCE_ID))
+                                .withFieldSettingsId(fieldSettingsByColumnNameRs.getInt(FIELD_SETTINGS_ID))
```
SimoneMaiwald, line 181: I think the order of field_settings_id and ddp_instance_id  in FieldSettingsDto is not correct here

src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
```diff
@@ -139,21 +140,20 @@ public int delete(int id) {
...
-                                        fieldSettingsByInstanceIdRs.getString(POSSIBLE_VALUES),
-                                        fieldSettingsByInstanceIdRs.getString(ACTIONS),
-                                        fieldSettingsByInstanceIdRs.getBoolean(READONLY),
-                                        fieldSettingsByInstanceIdRs.getInt(ORDER_NUMBER),
-                                        fieldSettingsByInstanceIdRs.getBoolean(DELETED),
-                                        fieldSettingsByInstanceIdRs.getLong(LAST_CHANGED),
-                                        fieldSettingsByInstanceIdRs.getString(CHANGED_BY)
-                                )
+                                new FieldSettingsDto.Builder(fieldSettingsByInstanceIdRs.getInt(DDP_INSTANCE_ID))
+                                        .withFieldSettingsId(fieldSettingsByInstanceIdRs.getInt(FIELD_SETTINGS_ID))
```
SimoneMaiwald, line 144: same here:  I think the order is different

src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
```diff
@@ -99,21 +101,20 @@ public int delete(int id) {
...
+                                        .withColumnDisplay(rs.getString(COLUMN_DISPLAY))
+                                        .withDisplayType(rs.getString(DISPLAY_TYPE))
+                                        .withPossibleValues(rs.getString(POSSIBLE_VALUES))
+                                        .withActions(rs.getString(ACTIONS))
+                                        .withReadOnly(rs.getBoolean(READONLY))
+                                        .withOrderNumber(rs.getInt(ORDER_NUMBER))
+                                        .withDeleted(rs.getBoolean(DELETED))
+                                        .withLastChanged(rs.getLong(LAST_CHANGED))
+                                        .withChangedBy(rs.getString(CHANGED_BY))
+                                .build()
```
SimoneMaiwald, line 117: seeing that this is reused, maybe we could put it into a method, so that we do not need to copy/paste?

src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
gmakharat: Builder pattern is used here, constructor of Builder takes `ddp_instance_id` and ordering for helper methods, which start `with` prefix, does not matter.

src/main/java/org/broadinstitute/dsm/model/participant/data/ParticipantData.java
```diff
@@ -74,24 +73,29 @@ public static NewParticipantData parseDto(@NonNull ParticipantDataDto participan
         return participantData;
     }
 
-    public Map<String, String> mergeParticipantData(@NonNull AddFamilyMemberPayload familyMemberPayload) {
+    public void setFamilyMemberData(@NonNull AddFamilyMemberPayload familyMemberPayload) {
         FamilyMemberDetails familyMemberData =
                 familyMemberPayload.getData().orElseThrow(() -> new NoSuchElementException("Family member data is not provided"));
         familyMemberData.setCollaboratorParticipantId(
                 familyMemberPayload.getRealm().orElse("").toUpperCase() +
```
SimoneMaiwald: there is a collaborator_id_prefix in ddp_instance table which could be used for this. For RGP it is `RGP` and for CMI studies it is `[NAME]Project` in case other studies want to use it

SimoneMaiwald: maybe worth changing, but we could also do that at a later point and leave it as it is for now

src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
gmakharat: I think main purpose of Builder pattern is that not to have such method, just because several `DAO` method uses all the columns from `field_settings` table does not worth to create such method in my opinion, later we may have queries like which may only return only 3 column(`ddp_instance_id`, `field_settings_id`, `column_name` for example), that method will not work on such cases and may produce confusion, that's why Builder is more handy, what do you think?

src/main/java/org/broadinstitute/dsm/model/participant/data/ParticipantData.java
gmakharat: I like that idea actually, since it will provide more dynamicity, will check if it is not big deal to do it quickly

src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
SimoneMaiwald: Oh I missed the Builder constructor having instance_id :doh:

src/main/java/org/broadinstitute/dsm/db/dao/fieldsettings/FieldSettingsDao.java
SimoneMaiwald: 👍

src/main/java/org/broadinstitute/dsm/model/participant/data/ParticipantData.java
SimoneMaiwald: 👍

src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/DDPInstanceDao.java
```diff
@@ -230,5 +235,29 @@ public int getDDPInstanceIdByGuid(@NonNull String studyGuid) {
...
+                }
+            }
+            catch (SQLException ex) {
+                dbVals.resultException = ex;
+            }
+            return dbVals;
+        });
 
+        if (results.resultException != null) {
+            throw new RuntimeException("Couldn't get participant es index with study guid: " + studyGuid, results.resultException);
```
SimoneMaiwald: Not the correct error message here

src/main/java/org/broadinstitute/dsm/db/dao/ddp/instance/DDPInstanceDao.java
```diff
@@ -230,5 +235,29 @@ public int getDDPInstanceIdByGuid(@NonNull String studyGuid) {
...
+            SimpleResult dbVals = new SimpleResult();
+            try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_COLLABORATOR_ID_PREFIX_BY_STUDY_GUID)) {
+                stmt.setString(1, studyGuid);
+                try (ResultSet rs = stmt.executeQuery()) {
+                    if (rs.next()) {
+                        dbVals.resultValue = rs.getString(COLLABORATOR_ID_PREFIX);
+                    }
+                }
+                catch (SQLException e) {
+                    throw new RuntimeException("Error getting participant es index with study guid: " + studyGuid, e);
```
SimoneMaiwald: and here

PR #224, opened 2021-07-07
Ddp 6508 return empty list when needed

Irakli-Khaladze:


pegahtah (approved):

src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
```diff
@@ -313,7 +315,7 @@ public static void addParticipantDataFilters(Map<String, String> queryConditions
             String data = participantData.getData();
             if (data != null) {
                 JsonObject dataJsonObject = gson.fromJson(data, JsonObject.class);
-                if (filter.getSelectedOptions() != null) {
+                if (OPTIONS.equals(filter.getType()) || RADIO.equals(filter.getType())) {
```
SimoneMaiwald: I think we should still add a != null check here. not to throw an NPE at L319

PR #231, opened 2021-07-08
Ddp 6507 non empty dates search fix

Irakli-Khaladze:
Added not empty field check and range check for dynamic form fields in addParticipantDataFilters method

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
+++ b/src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
@@ -57,6 +57,187 @@ public class FilterRoute extends RequestHandler {
         this.patchUtil = patchUtil;
     }
 
+    @Override
+    public Object processRequest(Request request, Response response, String userId) throws Exception {
+        if (patchUtil.getColumnNameMap() == null) {
+            throw new RuntimeException("ColumnNameMap is null!");
+        }
+        QueryParamsMap queryParams = request.queryMap();
+        String parent = null;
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
```diff
@@ -348,14 +343,38 @@ public static void addParticipantDataFilters(Map<String, String> queryConditions
...
+                                            AtomicBoolean participantAdded, AtomicBoolean first, ParticipantDataDto participantData,
+                                            JsonObject dataJsonObject) {
+        Object value1 = filter.getFilter1().getValue();
+        Object value2 = filter.getFilter2().getValue();
+        if (value1 == null || value2 == null) {
+            return;
+        }
+        boolean inNumberRange = value1 instanceof Double && value2 instanceof Double && dataJsonObject.get(tmpName) != null
+                && Double.compare(dataJsonObject.get(tmpName).getAsDouble(), (Double) value1) > 0
+                && Double.compare(dataJsonObject.get(tmpName).getAsDouble(), (Double) value2) < 0;
```
SimoneMaiwald: would it still be inNumberRange if value1 = 5 and value2 = 3

Irakli-Khaladze: Didn't know that I had to consider it from business logic perspective. I'll add to check that

SimoneMaiwald (approved):

PR #234, opened 2021-07-09
Ddp 6541 empty check for dynamic fields

Irakli-Khaladze:
Added empty check in filtering. Refactored and changed filtering logic a bit: first save participant id-s and then add them to the query and also check if it's guid or altpid

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
@@ -73,12 +73,12 @@ public class ElasticSearchUtil {
     public static final String PDFS = "pdfs";
     public static final String GUID = "guid";
     public static final String LEGACY_ALT_PID = "legacyAltPid";
-    public static final String BY_GUID = " AND profile.guid = ";
+    public static final String BY_GUID = "profile.guid = ";
     public static final String EMPTY = "";
     public static final String BY_HRUID = " AND profile.hruid = ";
     public static final String BY_GUIDS = " OR profile.guid = ";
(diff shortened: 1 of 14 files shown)


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -73,12 +73,12 @@
     public static final String PDFS = "pdfs";
     public static final String GUID = "guid";
     public static final String LEGACY_ALT_PID = "legacyAltPid";
-    public static final String BY_GUID = " AND profile.guid = ";
+    public static final String BY_GUID = "profile.guid = ";
```
SimoneMaiwald: That is used in a lot of places, are you sure you want to change it?

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -73,12 +73,12 @@
...
     public static final String EMPTY = "";
     public static final String BY_HRUID = " AND profile.hruid = ";
     public static final String BY_GUIDS = " OR profile.guid = ";
-    public static final String BY_LEGACY_ALTPID = " AND profile.legacyAltPid = ";
-    public static final String BY_LEGACY_ALTPID_STARTING = " AND (profile.legacyAltPid = ";
+    public static final String BY_LEGACY_ALTPID = "profile.legacyAltPid = ";
```
SimoneMaiwald: used in multiple classes, I don't think you should change that

SimoneMaiwald (approved):

PR #235, opened 2021-07-09
DDP-6238-PDF

pegahtah:
place holders for pdf


src/main/java/org/broadinstitute/dsm/db/dao/user/UserDao.java
```diff
@@ -31,7 +31,8 @@
                     if (rs.next()) {
                         dbVals.resultValue = new UserDto(rs.getInt(USER_ID),
                                 rs.getString(NAME),
-                                rs.getString(EMAIL));
+                                rs.getString(EMAIL),
+                                rs.getString("phone_number"));
```
gmakharat: extract String literal as a constant like other fields.

src/main/resources/master-changelog.xml
```diff
@@ -62,6 +62,7 @@
     <include file="liquibase/DDP-6006_RGP-form.xml" relativeToChangelogFile="true"/>
     <include file="liquibase/tab_grouped-column-display-fix.xml" relativeToChangelogFile="true"/>
     <include file="liquibase/DDP-6489-family-notes-field.xml" relativeToChangelogFile="true"/>
+    <include file="liquibase/DDP-6238-users_phoneNumbers.xml" relativeToChangelogFile="true"/>
```
gmakharat: we follow DDP-XXX_YY-YY.xml pattern of naming

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -76,7 +78,7 @@ public Object processRequest(Request request, Response response, String userId)
                     if (StringUtils.isNotBlank(requestBody)) {
                         JSONObject jsonObject = new JSONObject(requestBody);
                         String userIdR = (String) jsonObject.get(RequestParameter.USER_ID);
-                        Integer userIdRequest = Integer.parseInt(userIdR);
+                        Long userIdRequest = Long.parseLong(userIdR);
```
gmakharat: The comment is not related to this specific line but as I see this class is written exactly like `FilterRoute` was, maybe it is good time to refactor it, before it will become as messy as `FilterRoute` was?

pegahtah: I agree with the refactoring in general. I will put that on my list, not sure if we have to do it in this ticket or not since I think its good to have this tested and merged. 
@SimoneMaiwald what do you think?

gmakharat: most of the time postponed refactoring ends up with the hope that some day someone will refactor it and after a while we have doubled size messed up code than something related to refactoring.

SimoneMaiwald: yes, I think we should refactor it now, it is a good point now that we have bundle pdf and this change, I think it is time to do it and then also catch the not able to merge error better. with a nicer message in slack (maybe user and pt id and what pdf)

gmakharat: @pegahtah separation looks good, but route itself still looks more or less complicated, because of lots of if/if chaining, I would recommend to use Factory Object design pattern, factory class which would be responsible to generate different pdf classes according to some criteria, it would help to release route class from `if`s, also try to use guard condition approach instead of putting huge logic in if blocks, better approach of guard condition is `Optional` if it is possible to use, but since `Request` class of java is not designed to return `Optional` and instead it return `null`, guard conditions are more convenient.

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -70,434 +37,43 @@ public Object processRequest(Request request, Response response, String userId)
...
-                            return new Result(200);
-                        }
-                        else {
-                            response.status(500);
-                            throw new RuntimeException("Error missing ddpParticipantId");
-                        }
+                        Long userIdRequest = Long.parseLong((String) new JSONObject(requestBody).get(RequestParameter.USER_ID));
+                        DownloadPDF downloadPDFRequest = new DownloadPDF(requestBody);
+                        UserDto user = new UserDao().get(userIdRequest).orElseThrow();
+                        return downloadPDFRequest.getPDFs(response, user, realm, requestBody);
```
pegahtah: So the basic idea here is to separate downloading from participant page and getting the list of available pdfs for miscellaneous download. 
In order to do the first, I create a member of the class `DownloadPDF` and pass the request body to that class to decide what kind of pdf(s) to return

src/main/java/org/broadinstitute/dsm/util/UserUtil.java
```diff
@@ -317,7 +313,10 @@ private static void getList(Connection conn, String query, String role, String u
         }
     }
 
-    public static boolean checkUserAccess(String realm, String userId, String role) {
+    public boolean checkUserAccess(String realm, String userId, String role,  String userIdRequest) {
+        if (StringUtils.isNotBlank(userIdRequest) && !userId.equals(userIdRequest)) {
+            throw new RuntimeException("User id was not equal. User Id in token " + userId + " user Id in request " + userIdRequest);
```
pegahtah, line 318: Refactored this part from all the routes

src/main/java/org/broadinstitute/dsm/model/PDF/DownloadPDF.java
```diff
@@ -0,0 +1,291 @@
...
+                this.oncHistoryIDs.add((String) array.get(i));
+            }
+        }
+        if (jsonObject.has("pdfs")) {
+            this.pdfs =Arrays.asList(new Gson().fromJson(jsonObject.getString("pdfs"), PDF[].class));
+        }
+    }
+
+
+    public Object getPDFs(@NonNull Response response, UserDto user, String realm, String requestBody) {
```
gmakharat: Class needs to be as much independent as possible, passing `Response` object and manipulating from here does not look good, try to make all `Response` manipulations in `Route` classes.

src/main/java/org/broadinstitute/dsm/model/PDF/DownloadPDF.java
```diff
@@ -0,0 +1,291 @@
...
+                }
+                pdfMerger.mergeDocuments(MemoryUsageSetting.setupTempFileOnly());
+                //todo get page count and add them to the cover/request pdf
+            }
+            catch (IOException e) {
+                throw new FileProcessingException("Unable to merge documents ", e);
+            }
+            return output.toByteArray();
+        }
+        return null;
```
gmakharat: I would like to know why you always prefer returning null than returning `Optional`.

src/main/java/org/broadinstitute/dsm/model/PDF/DownloadPDF.java
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    protected byte[] generatePDFFromValues(Map<String, Object> valueMap, DDPInstance ddpInstance, PDFProcessor processor) {
+        InputStream stream = null;
```
gmakharat: `InputStream` has type of `AutoCloseable`, that's why you can use `try-with-resource` construction which is responsible to automatically close every resource, thus it gives you much cleaner code than this `finally` and closing resources manually.
```
try (InputStream stream = processor.generateStream(valueMap)) {
            stream.mark(0);
            return IOUtils.toByteArray(stream);
} catch (IOException e) {
            throw new RuntimeException("Couldn't get pdf for participant " + ddpParticipantId + " of ddpInstance " + ddpInstance.getName(), e);
}
```

src/main/java/org/broadinstitute/dsm/model/PDF/DownloadPDF.java
```diff
@@ -0,0 +1,291 @@
...
+            return DDPRequestUtil.getPDFByteArray(dsmRequest, ddpInstance.getName(), ddpInstance.isHasAuth0Token());
+        }
+        catch (IOException e) {
+            throw new RuntimeException("Couldn't get pdf for participant " + ddpParticipantId + " of ddpInstance " + ddpInstance.getName(), e);
+        }
+    }
+
+    public List<String> getPDFs(@NonNull String realm) {
+        List<String> listOfPDFs = new ArrayList<>();
+        SimpleResult results = inTransaction((conn) -> {
```
gmakharat: `DAO` here.

src/main/java/org/broadinstitute/dsm/model/PDF/MiscPDFDownload.java
```diff
@@ -0,0 +1,70 @@
...
+                if (dsm != null) {
+                    Object pdf = dsm.get(ElasticSearchUtil.PDFS);
+                    return pdf;
+                }
+            }
+            else {
+                for (Object value : participantESData.values()) {
+                    participantData = (Map<String, Object>) value;
+                    //check that it is really right participant
+                    if (participantData != null) {
```
gmakharat, line 52: too many if chain, guard condition approach could provide better visibility, or at least this whole if chains could be extracted as a method with some meaningful name.

src/main/java/org/broadinstitute/dsm/model/PDF/DownloadPDF.java
pegahtah: Honestly I don't see that big of a difference in the usage, both throw an exception when null and both need and if else check. 
However I changed this so the whole code will converge in style

src/main/java/org/broadinstitute/dsm/model/PDF/DownloadPDF.java
pegahtah: Good point!

src/main/java/org/broadinstitute/dsm/model/PDF/DownloadPDF.java
gmakharat: @pegahtah, first class is **without optional**, 
there are two examples to work with null:
`getInsuranceNameExample1` - example 1, used widely in legacy projects
`getInsuranceNameExample2` - `guard condition` approach, more or less good, better than first example, but drawback of this approach is that we have many return points which is not good for debugging and etc.
```
public class WithoutOptional {

	
	
	public static void main(String[] args) {
		
		
		Person p = new Person();
		
		String name = getInsuranceNameExample1(p);
		
		String name2 = getInsuranceNameExample2(p);
		
	}

	private static String getInsuranceNameExample2(Person p) {
		Objects.requireNonNull(p);
		Car car = p.getCar();
		if (car == null) return "Unknown";
		Insurance ins = car.getInsurance();
		if (ins == null) return "Unknown";
		String insName = ins.getName();
		if (insName == null) return "Unkown";
		return insName;
	}

	private static String getInsuranceNameExample1(Person p) {
		Objects.requireNonNull(p);
		Car car = p.getCar();
		if (car != null) {
			Insurance ins = car.getInsurance();
			if (ins != null) {
				String insName = ins.getName();
				if (insName != null) {
					return insName;
				}
			}
		}
		return "Unknown";
	}
	
}


class Person {
	private Car car;
	
	public Car getCar() {
		return car;
	}
	
	public void setCar(Car car) {
		this.car = car;
	}
	
}

class Insurance {
	private String name;
	
	public Insurance(String name) {
		this.name = name;
	}
	
	public String getName() {
		return name;
	}
}


class Car {
	private Insurance insurance;
	
	public Insurance getInsurance() {
		return insurance;
	}
	
	public void setInsurance(Insurance insurance) {
		this.insurance = insurance;
	}
}
```
second class, **with optional**:
`getInsuranceName` method to compare `getInsuranceNameExample1` and `getInsuranceNameExample2`, both do the same.
```
public class WithOptional {

	
	
	public static void main(String[] args) {
		
		
		Insurance tbc = new Insurance();
		tbc.setName("TBC");
		
		Car car = new Car();
		car.setInsurance(tbc);
		
		Person p = new Person();
		
		p.setCar(car);
		
		String name = getInsuranceName(p);
		
		System.out.println(name);
		
		
	}

	private static String getInsuranceName(Person p) {
		Objects.requireNonNull(p);
		return p.getCar()
			.flatMap(car -> car.getInsurance())
			.flatMap(ins -> ins.getName())
			.orElse("Unknown");		
	}
	
}


class Person {

	
	private Car car;
	
	public Optional<Car> getCar() {
		return Optional.ofNullable(car);
	}
	
	public void setCar(Car car) {
		this.car = car;
	}
	
	
}

class Car {

	
	private Insurance insurance;
	
	public Optional<Insurance> getInsurance() {
		return Optional.ofNullable(insurance);
	}
	
	public void setInsurance(Insurance ins) {
		this.insurance = ins;
	}
	
}



class Insurance {

	
	private String name;
	
	public Optional<String> getName() {
		return Optional.ofNullable(name);
	}
	
	public void setName(String name) {
		this.name = name;
	}
	
}
```

src/main/java/org/broadinstitute/dsm/route/DownloadPDFRoute.java
```diff
@@ -40,55 +38,46 @@ public Object processRequest(Request request, Response response, String userId)
...
-                    MiscPDFDownload miscPDFDownload = new MiscPDFDownload();
-                    String ddpParticipantId = null;
-                    if (queryParams.value(RequestParameter.DDP_PARTICIPANT_ID) != null) {
-                        ddpParticipantId = queryParams.get(RequestParameter.DDP_PARTICIPANT_ID).value();
+                Long userIdRequest = Long.parseLong((String) new JSONObject(requestBody).get(RequestParameter.USER_ID));
+                DownloadPDF downloadPDFRequest = new DownloadPDF(requestBody);
+                Optional<byte[]> pdfBytes = downloadPDFRequest.getPDFs(userIdRequest, realm, requestBody);
+                pdfBytes.ifPresent(pdfBytesArray -> {
+                    try {
+                        HttpServletResponse rawResponse = response.raw();
```
gmakharat, line 56: why it became necessary to use raw response and flushing result manually?

src/main/java/org/broadinstitute/dsm/model/PDF/MRCoverPDF.java
```diff
@@ -0,0 +1,89 @@
...
+        String startDate = null;
+        String endDate = null;
+        if (keySet.contains(JSON_START_DATE)) {
+            startDate = (String) jsonObject.get(JSON_START_DATE);
+            if (!"0/0".equals(startDate) && !startDate.contains("/") && startDate.contains("-")) {
+                startDate = SystemUtil.changeDateFormat(SystemUtil.DATE_FORMAT, SystemUtil.US_DATE_FORMAT, startDate);
+            }
+            else if (StringUtils.isNotBlank(startDate) && startDate.startsWith("0/") && !startDate.equals("0/0")) {
+                startDate = "01/" + startDate.split("/")[1];
+            }
```
SimoneMaiwald, line 41: I think this if statement is the same as in DownloadPDF dob check, maybe move it into a method?

src/main/java/org/broadinstitute/dsm/model/PDF/MRCoverPDF.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+        //get information from db
+        MedicalRecord medicalRecord = MedicalRecord.getMedicalRecord(ddpInstance.getName(), originalDownloadPDF.getDdpParticipantId(), originalDownloadPDF.getMedicalRecordId());
+
+        Map<String, Object> valueMap = new HashMap<>();
+        //values same no matter from where participant/institution data comes from
+        valueMap.put(CoverPDFProcessor.FIELD_CONFIRMED_INSTITUTION_NAME, medicalRecord.getName());
+        valueMap.put(CoverPDFProcessor.FIELD_CONFIRMED_INSTITUTION_NAME_2, medicalRecord.getName() + ",");
+        valueMap.put(CoverPDFProcessor.FIELD_CONFIRMED_FAX, medicalRecord.getFax());
+        String today = new SimpleDateFormat("MM/dd/yyyy").format(new Date());
```
SimoneMaiwald, line 58: for the today string see if we already have a method, if not move the date format string to static and up, so it is easier to find it if we need to make changes

src/main/java/org/broadinstitute/dsm/model/PDF/TissueCoverPDF.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public TissueCoverPDF(@NonNull DownloadPDF downloadPDF){
+        this.originalDownloadPDF = downloadPDF;
+    }
+
+    public Map<String, Object> getValuesForTissueCover(DDPInstance ddpInstance, UserDto user) {
+
+        logger.info("Generating request pdf for onc history ids {}", StringUtils.join(originalDownloadPDF.oncHistoryIDs, ","));
+        Map<String, Object> valueMap = new HashMap<>();
+        String today = new SimpleDateFormat("MM/dd/yyyy").format(new Date());
```
SimoneMaiwald, line 29: again, can we move the format string, maybe use the same in all pdf classes or from system utils or utility class

src/main/java/org/broadinstitute/dsm/route/AbstractionFormControlRoute.java
```diff
@@ -31,8 +31,9 @@ public Object processRequest(Request request, Response response, String userId)
         }
 
         if (StringUtils.isNotBlank(realm)) {
+            UserUtil userUtil = new UserUtil();
             if (RoutePath.RequestMethod.GET.toString().equals(request.requestMethod())) {
-                if (UserUtil.checkUserAccess(realm, userId, "mr_view")) {
+                if (userUtil.checkUserAccess(realm, userId, "mr_view", null)) {
```
SimoneMaiwald: if you are already touch all classes to change this, can you also change the role string and get it from a nice central point ("mr_view", "mr_request" and all of them) to minimize typos in the check. thank you

src/main/java/org/broadinstitute/dsm/util/DDPRequestUtil.java
```diff
@@ -305,22 +305,7 @@ public static void makeStandardPDF(@NonNull boolean hasConsentEndpoints, @NonNul
...
-            participantESData = ElasticSearchUtil.getFilteredDDPParticipantsFromES(ddpInstance, ElasticSearchUtil.BY_LEGACY_ALTPID + ddpParticipantId);
-            if (participantESData != null && !participantESData.isEmpty()) {
-                makeNonStandardPDF(participantESData, ddpInstance, ddpParticipantId, userId, reason);
-            }
-        }
-    }
-
-    private static void makeNonStandardPDF(@NonNull Map<String, Map<String, Object>> participantESData, @NonNull DDPInstance ddpInstance, @NonNull String ddpParticipantId, @NonNull String userId, @NonNull String reason) {
-        Object pdfs = DownloadPDFRoute.returnPDFS(participantESData, ddpParticipantId);
+        Object pdfs = new MiscPDFDownload().returnPDFS(ddpParticipantId, ddpInstance.getName());
```
SimoneMaiwald, line 308: nice 👍

src/main/java/org/broadinstitute/dsm/util/UserUtil.java
```diff
@@ -221,9 +217,9 @@ public static String getUserId(Request request) {
                     getList(conn, query, instanceRole, userId, listOfRealms);
                 }
                 else if (MEDICALRECORD_MENU.equals(menu)) {
-//                    query = query + QueryExtension.BY_ROLE_NAME;
-//                    query = query.replace("%1", DBConstants.MR_VIEW);
-//                    getList(conn, query, instanceRole, userId, listOfRealms);
+                    //                    query = query + QueryExtension.BY_ROLE_NAME;
+                    //                    query = query.replace("%1", DBConstants.MR_VIEW);
+                    //                    getList(conn, query, instanceRole, userId, listOfRealms);
```
SimoneMaiwald, line 222: just remove that

src/main/java/org/broadinstitute/dsm/route/BasicRoute.java
```diff
@@ -0,0 +1,37 @@
...
+        logger.info(request.url());
+        QueryParamsMap queryParams = request.queryMap();
+        String realm = null;
+        if (queryParams.value(RoutePath.REALM) != null) {
+            realm = queryParams.get(RoutePath.REALM).value();
+        }
+        UserUtil userUtil = new UserUtil();
+        String userIdR = userUtil.getUserId(request);
+        if (!userUtil.checkUserAccess(realm, userId, accessRole, userIdR)) {
+            return 404;
```
SimoneMaiwald: what about: 

```
response.status(404);
return null;
```

PR #239, opened 2021-07-12
DDP-6504 | sample information tab

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-6504

**Things have done:**

- Separated liquibase directory for rgp.
- Added liquibase file to add `hide_samples_tab` boolean column, set `hide_samples_tab` to true for `RGP`.
- Created DAO/DTO class for InstanceSettings.
- Refactored(not fully) existing `InstanceSetting` class.


src/main/resources/liquibase/rgp/DDP-6504_hide-sample-information-tab.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SimoneMaiwald, line 1: nice! I like the new study folder 👍

SimoneMaiwald (approved):

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -84,7 +85,10 @@ public void accept(PubsubMessage pubsubMessage, Context context) throws Exceptio
                     int lastKitId = 0;
                     UPSKit kit = null;
                     logger.info("tracking ups ids for " + ddpInstance.getName());
-                    boolean gbfShippedTriggerDSSDelivered = InstanceSettings.getInstanceSettings(ddpInstance.getName(), conn).isGbfShippedTriggerDSSDelivered();
+                    InstanceSettingsDto instanceSettings = new InstanceSettings().getInstanceSettings(ddpInstance.getName());
```
pegahtah: I think that won't work for CF. you need to pass the same connection so it knows how to connect to DB and get the instance settings

src/main/java/org/broadinstitute/dsm/db/InstanceSettings.java
```diff
@@ -68,59 +77,22 @@ public InstanceSettings(List<Value> mrCoverPdf, List<Value> kitBehaviorChange, L
...
-        });
+    public InstanceSettings() {
 
-        if (results.resultException != null) {
-            throw new RuntimeException("Error getting list of realms ", results.resultException);
-        }
-        return (InstanceSettings) results.resultValue;
     }
 
-    public static InstanceSettings getInstanceSettings(@NonNull String realm, Connection conn) {
```
pegahtah, line 101: This is what was used before @gmakharat for reference

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
gmakharat: @pegahtah does it use different database or?

pegahtah (changes requested):
I am going to stop this from being merged by submitting the `request changes` feedback. So it won't get merged accidentally and mess with CF

pegahtah (approved):

PR #245, opened 2021-07-14
Creating Clinical Kit endpoints

pegahtah:


src/main/java/org/broadinstitute/dsm/DSMServer.java
```diff
@@ -209,6 +210,8 @@ protected void setupCustomRouting(@NonNull Config cfg) {
         get(API_ROOT + RoutePath.BSP_KIT_REGISTERED, new BSPKitRegisteredRoute(), new JsonTransformer());
         if(!cfg.getBoolean("ui.production")){
             get(API_ROOT + RoutePath.DUMMY_ENDPOINT, new MercuryDummyKitEndpoint(), new JsonTransformer());
+            get(API_ROOT + RoutePath.CLINICAL_KIT_ENDPOINT, new ClinicalKitsRoute(notificationUtil), new JsonTransformer());
```
SimoneMaiwald: this is a real endpoint we need in prod (later on, so I would already move it to where the other gp route is

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,79 @@
...
+                kitInfo.getSampleCollectionBarcode(),
+                kitInfo.getMaterialInfo(),
+                kitInfo.getReceptacleName(),
+                null,
+                null);
+
+        BSPKitQueryResult bspKitQueryResult = BSPKitQueryResult.getBSPKitQueryResult(kitLabel);
+        DDPInstance ddpInstance = DDPInstance.getDDPInstance(bspKitQueryResult.getInstanceName());
+
+        Optional<ParticipantWrapper> maybeParticipant = ParticipantWrapper.getParticipantFromESByHruid(ddpInstance, bspKitQueryResult.getBspParticipantId());
```
SimoneMaiwald: That will not work! bspParticipantId has the project prefix.
you need to search with GUID

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,79 @@
...
+        return TransactionWrapper.inTransaction(conn -> {
+            return getClinicalKit(conn, kitLabel, response);
+        });
+    }
+
+    private Object getClinicalKit(Connection conn, String kitLabel, Response response) {
+        logger.info("Checking label " + kitLabel);
+
+        // this method already sets the received time, check for exited and deactivation and special behaviour, and triggers DDP, we don't need a new
+        BSPKitInfo kitInfo = (BSPKitInfo) BSPKitQueryRoute.getBSPKitInfo(conn, kitLabel, response, notificationUtil);
```
SimoneMaiwald: don't use another route in a route! move shared methods into a shared class (util class or other)

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,79 @@
...
+
+        Optional<ParticipantWrapper> maybeParticipant = ParticipantWrapper.getParticipantFromESByHruid(ddpInstance, bspKitQueryResult.getBspParticipantId());
+        maybeParticipant.ifPresent(p -> {
+            Map<String, String> dsm = (Map<String, String>) p.getData().get("dsm");
+                if (dsm != null && !dsm.isEmpty()) {
+                    clinicalKit.setDateOfBirth((String) dsm.get("dateOfBirth"));
+                }
+            Map<String, String> participantProfileFromEs = (Map<String, String>) p.getData().get("profile");
+                String firstName =  participantProfileFromEs.get("firstName");
+                String lastName =  participantProfileFromEs.get("lastName");
```
SimoneMaiwald: don't use these strings directly

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -0,0 +1,82 @@
...
+    private static String CLINICAL_KIT_REALM = "CLINICAL_KIT_REALM";
+    private static int REALM;
+    private static final String USER_ID = "74";
+
+    @Override
+    public Object handle(Request request, Response response) throws Exception {
+        String kitLabel = request.params(RequestParameter.LABEL);
+        if (StringUtils.isBlank(kitLabel)) {
+            logger.warn("Got a create Clinical Kit request without a kit label!!");
+            return new Result(400, "Please include a kit label as a path parameter");
```
SimoneMaiwald: I think we should just do a response.status(500) here

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -0,0 +1,82 @@
...
+import spark.Response;
+import spark.Route;
+
+import java.util.List;
+import java.util.Map;
+
+public class CreateClinicalDummyKitRoute implements Route {
+    private static final Logger logger = LoggerFactory.getLogger(CreateClinicalDummyKitRoute.class);
+    private static String CLINICAL_KIT_REALM = "CLINICAL_KIT_REALM";
+    private static int REALM;
```
SimoneMaiwald: why static?

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -0,0 +1,82 @@
...
+                if (participantESData != null && !participantESData.isEmpty()) {
+                    Map<String, Object> profile = (Map<String, Object>) participantESData.get("profile");
+                    String participantCollaboratorId = KitRequestShipping.getCollaboratorParticipantId(ddpInstance.getBaseUrl(), ddpInstance.getDdpInstanceId(), ddpInstance.isMigratedDDP(),
+                            ddpInstance.getCollaboratorIdPrefix(), ddpParticipantId, (String) profile.get("hruid"), null);
+                    String collaboratorSampleId = MercuryDummyKitEndpoint.getCollaboratorSampleId(salivaKitType.getKitId(), participantCollaboratorId);
+                    if (ddpParticipantId != null) {
+                        //if instance not null
+                        String dsmKitRequestId = KitRequestShipping.writeRequest(ddpInstance.getDdpInstanceId(), kitRequestId, salivaKitType.getKitId(),
+                                ddpParticipantId, participantCollaboratorId, collaboratorSampleId,
+                                USER_ID, "", "", "", false, "");
```
SimoneMaiwald: don't user USER_ID for created_by use "MERCURY"

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -0,0 +1,82 @@
...
+                    logger.info("Kit added successfully");
+                    return new Result(200);
+                }
+                logger.error("Participants for realm " + ddpInstance.getName() + " not found in ES");
+
+            }
+            logger.error("Participant " + ddpParticipantId + " was not found in ES!");
+        }
+        logger.error("Error occurred while adding kit");
+        return new Result(500);
```
SimoneMaiwald: I think we should change that to response.status(500)

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -0,0 +1,82 @@
...
+                    String collaboratorSampleId = MercuryDummyKitEndpoint.getCollaboratorSampleId(salivaKitType.getKitId(), participantCollaboratorId);
+                    if (ddpParticipantId != null) {
+                        //if instance not null
+                        String dsmKitRequestId = KitRequestShipping.writeRequest(ddpInstance.getDdpInstanceId(), kitRequestId, salivaKitType.getKitId(),
+                                ddpParticipantId, participantCollaboratorId, collaboratorSampleId,
+                                USER_ID, "", "", "", false, "");
+                        MercuryDummyKitEndpoint.updateKitLabel(kitLabel, dsmKitRequestId);
+                    }
+                    logger.info("Kit added successfully");
+                    return new Result(200);
```
SimoneMaiwald: response.status(200) and not returning Result object

src/main/java/org/broadinstitute/dsm/route/MercuryDummyKitEndpoint.java
```diff
@@ -65,7 +65,7 @@ public Object handle(Request request, Response response) throws Exception {
         return new Result(500);
```
SimoneMaiwald: I think we should rename the old class to CreateBSPDummyKitRoute

src/main/java/org/broadinstitute/dsm/route/BSPKitQueryRoute.java
```diff
@@ -46,11 +45,11 @@ public Object handle(Request request, Response response) throws Exception {
         }
 
         return TransactionWrapper.inTransaction(conn -> {
-            return getBSPKitInfo(conn, kitLabel, response);
+            return getBSPKitInfo(conn, kitLabel, response, this.notificationUtil);
```
gmakharat: having DB connection directly in route is not a good idea, new `DAO` class for it would be good.

src/main/java/org/broadinstitute/dsm/route/BSPKitQueryRoute.java
```diff
@@ -46,11 +45,11 @@ public Object handle(Request request, Response response) throws Exception {
...
         });
     }
 
-    public Object getBSPKitInfo(Connection conn, @NonNull String kitLabel, @NonNull Response response) {
+    public static Object getBSPKitInfo(Connection conn, @NonNull String kitLabel, @NonNull Response response, NotificationUtil notificationUtil) {
```
gmakharat: making method `static` adds more complexity, for example, `notificationUtil` is instance field and you had to add it as a parameter which automatically adds more complexity, introducing `DAO` class would omit passing `Connection` to this method, method has several return points and it may return different type of results which adds more confusion, dissembling method into more specific parts would bring more cleaner code, except that, I always prefer to use method result for decisions, for example, you could use `null` result in upper method(`handle`) to decide whether make response status 400 or not, this approach would help to remove one more argument from method(Response response).

src/main/java/org/broadinstitute/dsm/route/BSPKitQueryRoute.java
```diff
@@ -138,15 +137,15 @@ public Object getBSPKitInfo(Connection conn, @NonNull String kitLabel, @NonNull
 
     }
 
-    private void writeSampleReceivedToES(DDPInstance ddpInstance, BSPKitQueryResult bspKitInfo) {
+    private static void writeSampleReceivedToES(DDPInstance ddpInstance, BSPKitQueryResult bspKitInfo) {
```
gmakharat: result of making instance methods static, Java is designed to be used for object oriented programming, static methods usually breaks all the principles of OOP, in OOP static methods are usually used as utility methods, which calculates something and returns new result instead of mutation.

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
gmakharat: I would add here that, since you use some logic from `BSPKitQueryRoute` you could introduce new class for example `BSPKitQuery` and move all necessary logic there and you could freely use object of `BSPKitQuery` in both route `ClinicalKitsRoute` and `BSPKitQueryRoute`.

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,79 @@
...
+        if (StringUtils.isBlank(kitLabel)) {
+            throw new RuntimeException("Please include a kit label as a path parameter");
+        }
+
+        return TransactionWrapper.inTransaction(conn -> {
+            return getClinicalKit(conn, kitLabel, response);
+        });
+    }
+
+    private Object getClinicalKit(Connection conn, String kitLabel, Response response) {
```
gmakharat: It is always preferable to make methods more specific to keep high cohesion, as it seems, this method has only one return point and it returns object of type `ClinicalKit`, so it can be easily changed from `Object` to `ClinicalKit`, such small change would bring more understanding of method.

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
gmakharat: Since you have `ddpInstance` I would say this method `getParticipantESDataByParticipantId` from ElasticSearchUtil would be more preferable to get participant data, since you use only `dsm` and `profile` data from ES, this method `getParticipantFromESByHruid` has unnecessary complexity for your purpose.

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,79 @@
...
+                kitInfo.getMaterialInfo(),
+                kitInfo.getReceptacleName(),
+                null,
+                null);
+
+        BSPKitQueryResult bspKitQueryResult = BSPKitQueryResult.getBSPKitQueryResult(kitLabel);
+        DDPInstance ddpInstance = DDPInstance.getDDPInstance(bspKitQueryResult.getInstanceName());
+
+        Optional<ParticipantWrapper> maybeParticipant = ParticipantWrapper.getParticipantFromESByHruid(ddpInstance, bspKitQueryResult.getBspParticipantId());
+        maybeParticipant.ifPresent(p -> {
```
gmakharat: Since you use `orElseThrow` after this, Optional's `ifPresentOrElse` would be better to use here.

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -0,0 +1,82 @@
...
+                    break;
+                }
+            }
+            if (salivaKitType != null) {
+                logger.info("Got instance  " + ddpInstance.getName() + " and kit type id " + salivaKitType.getKitId());
+            }
+            String kitRequestId = "CLINICALKIT_" + KitRequestShipping.createRandom(20);
+
+            String ddpParticipantId = MercuryDummyKitEndpoint.getRandomParticipantIdForStudy(ddpInstance.getDdpInstanceId());
+            Map<String, Map<String, Object>> participantsESData = ElasticSearchUtil.getFilteredDDPParticipantsFromES(ddpInstance, ElasticSearchUtil.BY_GUID + ddpParticipantId);
```
gmakharat: `getFilteredDDPParticipantsFromES` method fetches all the participants information which I believe you don't need at all, since you have already `ddpParticipantId` you could use `getParticipantESDataByParticipantId` method instead of it, which would take much less time, also this method returns Optional which would help you to provide much cleaner code instead of this several if chaining.

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -0,0 +1,82 @@
...
+            return new Result(400, "Please include a kit label as a path parameter");
+        }
+        logger.info("Got a new Clinical Kit request with kit label " + kitLabel);
+        REALM = (int) DBUtil.getBookmark(CLINICAL_KIT_REALM);
+        KitType salivaKitType = null;
+        DDPInstance ddpInstance = DDPInstance.getDDPInstanceById(REALM);
+        if (ddpInstance != null) {
+            List<KitType> kitTypeIds = KitType.getKitTypes(ddpInstance.getName(), null);
+
+            for (KitType kitType : kitTypeIds) {
```
gmakharat: salivaKitType = `kitTypeIds.stream.filter(kitType -> "SALIVA".equals(kitType.getName())).findFirst().orElseThrow()`, I think this approach is much more concise and protected from unexpected later NPEs.

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
pegahtah: last time James wanted more info in the header so I added a message, but I will change the code to `500`

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,231 @@
...
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.*;
+
+import static org.broadinstitute.ddp.db.TransactionWrapper.inTransaction;
+
+public class BSPKitDao implements Dao<BSPKitDao> {
```
gmakharat: Dao needs to return Dto(or Dto like class which holds data fetched from DB) not itself.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,231 @@
...
+
+    private static final String SQL_UPDATE_DUMMY_KIT = "UPDATE ddp_kit SET kit_label = ? where dsm_kit_request_id = ?";
+    private static final String SQL_SELECT_RANDOM_PT = "SELECT ddp_participant_id FROM ddp_kit_request where ddp_instance_id = ?  ORDER BY RAND() LIMIT 1";
+    private static final Logger logger = LoggerFactory.getLogger(BSPKitDao.class);
+
+    public static void updateKitLabel(String kitLabel, String dsmKitRequestId) {
```
gmakharat: try to never use static, maybe from first sight it looks simpler than instance methods but in overall it once again breaks OOP.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,231 @@
...
+                dbVals.resultException = e;
+            }
+            return dbVals;
+        });
+        if (results.resultException != null) {
+            throw new RuntimeException("Error updating kit  label for " + dsmKitRequestId, results.resultException);
+        }
+    }
+
+    public static String getRandomParticipantIdForStudy(String ddpInstanceId) {
```
gmakharat: the same about static, additionally if method returns null reference it is always good to use Optionals, instead of returning `null` on line 84, you could return `Optional.ofNullable((String) results.resultValue)`, Optionals will help you to avoid future NPE's, someone who will decide to use this method he/she will have to work with Optional.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,231 @@
...
+        if (results.resultException != null) {
+            throw new RuntimeException("Problem getting a random participant id for instance id " + ddpInstanceId, results.resultException);
+        }
+        if (results.resultValue != null) {
+            return (String) results.resultValue;
+        }
+        return null;
+    }
+
+    public static String getCollaboratorSampleId(int kitTypeId, String participantCollaboratorId, String kitType) {
```
gmakharat: this method is not related to `BSPKitDao` at all, in general, as I see this class combines lots of different database tables, each Dao should be very specific, it should try maximally to serve only one table if it is possible, `getCollaboratorSampleId` is a business logic I would say, business logic should have its own class which will use `Dao` classes to make its job.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,231 @@
...
+        });
+        if (results.resultException != null) {
+            throw new RuntimeException("Error getting Collaborator Sample Id for  " + participantCollaboratorId, results.resultException);
+        }
+        else {
+            return (String) results.resultValue;
+        }
+    }
+
+    public Object getBSPKitInfo(@NonNull String kitLabel, @NonNull Response response, NotificationUtil notificationUtil) {
```
gmakharat: method wrapped in `Dao` class which work with `Response` as well, too many responsibilities for one class and method.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,231 @@
...
+                    bspSampleId,
+                    bspMaterialType,
+                    bspReceptacleType);
+        }
+        else {
+            throw new RuntimeException("No participant id for " + kitLabel + " from " + bspKitQueryResult.getInstanceName());
+        }
+    }
+
+    private static void writeSampleReceivedToES(DDPInstance ddpInstance, BSPKitQueryResult bspKitInfo) {
```
gmakharat: Not related to `Dao` at all, it does business logic.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,231 @@
...
+        String kitRequestId = new KitRequestDao().getKitRequestIdByBSPParticipantId(bspKitInfo.getBspParticipantId());
+        Map<String, Object> nameValuesMap = new HashMap<>();
+        nameValuesMap.put(ESObjectConstants.RECEIVED, SystemUtil.getISO8601DateString());
+        ElasticSearchUtil.writeSample(ddpInstance, kitRequestId, bspKitInfo.getDdpParticipantId(), ESObjectConstants.SAMPLES,
+                ESObjectConstants.KIT_REQUEST_ID, nameValuesMap);
+    }
+
+    private static void triggerDDP(Connection conn, @NonNull BSPKitQueryResult bspKitInfo, boolean firstTimeReceived, String kitLabel) {
```
gmakharat: the class is everything except `Dao`. https://en.wikipedia.org/wiki/Data_access_object

src/main/java/org/broadinstitute/dsm/route/CreateBSPDummyKitRoute.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(CreateBSPDummyKitRoute.class);
+
+    @Override
+    public Object handle(Request request, Response response) throws Exception {
+        logger.info("Got Mercury Test request");
+        String kitLabel = request.params(RequestParameter.LABEL);
+        if (StringUtils.isBlank(kitLabel)) {
+            response.status(400);// return bad request
+            logger.error("Bad request from Mercury! Should include a kitlabel");
+            return new Result(400, "Please include a kit label as a path parameter");
```
gmakharat: let's step by step move from using `Result` to using `Response` itself, if you want to return message you can simple return `String`.

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public Object handle(Request request, Response response) {
+        String kitLabel = request.params(RequestParameter.LABEL);
+        if (StringUtils.isBlank(kitLabel)) {
+            logger.warn("Got a create Clinical Kit request without a kit label!!");
+            return new Result(500, "Please include a kit label as a path parameter");
+        }
+        logger.info("Got a new Clinical Kit request with kit label " + kitLabel);
+        REALM = (int) DBUtil.getBookmark(CLINICAL_KIT_REALM);
```
gmakharat: there is `BookmarkDao` class which you can use methods from there or add new, if they don't satisfy your requirements.

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,87 @@
...
+        });
+    }
+
+    private ClinicalKitDto getClinicalKit(Connection conn, String kitLabel, Response response) {
+        logger.info("Checking label " + kitLabel);
+
+        // this method already sets the received time, check for exited and deactivation and special behaviour, and triggers DDP, we don't need a new
+        BSPKitInfo kitInfo = (BSPKitInfo) new BSPKitDao().getBSPKitInfo(kitLabel, response, notificationUtil);
+
+        ClinicalKitDto clinicalKit = new ClinicalKitDto(kitInfo.getCollaboratorParticipantId(),
```
gmakharat: having `Builder` pattern in `ClinicalKitDto` would be useful, since you don't use all of `ClinicalKitDto` attributes.

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public Object handle(Request request, Response response) {
+        String kitLabel = request.params(RequestParameter.LABEL);
+        if (StringUtils.isBlank(kitLabel)) {
+            throw new RuntimeException("Please include a kit label as a path parameter");
+        }
+
+        return TransactionWrapper.inTransaction(conn -> {
```
gmakharat: Dao should do this not route directly.

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,87 @@
...
+            return getClinicalKit(conn, kitLabel, response);
+        });
+    }
+
+    private ClinicalKitDto getClinicalKit(Connection conn, String kitLabel, Response response) {
```
gmakharat: it is not necessary to pass response here, you could manipulate response in upper level, in `handle` method according to the returned result of this method.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,85 @@
...
+                dbVals.resultException = e;
+            }
+            return dbVals;
+        });
+        if (results.resultException != null) {
+            throw new RuntimeException("Error updating kit  label for " + dsmKitRequestId, results.resultException);
+        }
+    }
+
+    public String getRandomParticipantIdForStudy(String ddpInstanceId) {
```
gmakharat: return Optional instead of string and null reference.
`public Optional<String> getRandomParticipantIdForStudy(String ddpInstanceId)`

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,85 @@
...
+            }
+            return dbVals;
+        });
+        if (results.resultException != null) {
+            throw new RuntimeException("Problem getting a random participant id for instance id " + ddpInstanceId, results.resultException);
+        }
+        if (results.resultValue != null) {
+            return (String) results.resultValue;
+        }
+        return null;
```
gmakharat: `return Optional.ofNullable((String) results.resultValue)`, if statement on line 65 won't be necessary after this.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,150 @@
...
+
+        if (StringUtils.isNotBlank(bspKitQueryResult.getParticipantExitId())) {
+            String message = "Kit of exited participant " + bspKitQueryResult.getBspParticipantId() + " was received by GP.<br>";
+            notificationUtil.sentNotification(bspKitQueryResult.getNotificationRecipient(), message, NotificationUtil.DSM_SUBJECT);
+            return new BSPKitStatus(BSPKitStatus.EXITED);
+        }
+        if (StringUtils.isNotBlank(bspKitQueryResult.getDeactivationDate())) {
+            return new BSPKitStatus(BSPKitStatus.DEACTIVATED);
+        }
+        return null;
```
gmakharat: main intent why Java introduced Optional from version 8 is to handle NPEs, main usage of optional is the methods where there is a chance of return of null reference. Additionally Optional not only provides handle of NPEs, with Optional code becomes much more cleaner since you don't have to write all those `null` reference if checks.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,150 @@
...
+import java.sql.Connection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class BSPKit {
+    private static Logger logger = LoggerFactory.getLogger(BSPKit.class);
+
+    public BSPKitStatus getKitStatus(@NonNull String kitLabel, NotificationUtil notificationUtil) {
+        logger.info("Checking label " + kitLabel);
+        BSPKitQueryResult bspKitQueryResult = BSPKitQueryResult.getBSPKitQueryResult(kitLabel);
```
gmakharat: `BSPKitQueryResult` class can be easily separated into `Dao/Dto` since it has only one method `getBSPKitQueryResult`, Optional can be applied on  `getBSPKitQueryResult` as well, which will help you to make later if check more cleaner.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.sql.Connection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class BSPKit {
+    private static Logger logger = LoggerFactory.getLogger(BSPKit.class);
+
+    public BSPKitStatus getKitStatus(@NonNull String kitLabel, NotificationUtil notificationUtil) {
+        logger.info("Checking label " + kitLabel);
```
gmakharat, line 32: personally I prefer to avoid many return points in method, what I would do, if `getBSPKitQueryResult` would return Optional I would use `map` on it, in `map` I would wrap all the logic from L35 to L43, in the beginning of lambda function of `map` I would declare `Optional<BSPKitStatus> result = Optional.empty()`, then in if statements I would change value of `result` accordingly `result = Optional.of(new BSPKitStatus(BSPKitStatus.EXITED))` and at the end of lambda function of map would return `result`, map invokes only if Optional has value inside it, if Optional returned by `getBSPKitQueryResult` would be empty I would apply `orElseGet` since you use logger as well after `map`, `orElseGet(() -> logger.info("text here"); return Optional.empty())`, that's how I would decrease number of return points, but as I mentioned in the beginning it is personal choice.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+        if (StringUtils.isNotBlank(bspKitQueryResult.getParticipantExitId()) || StringUtils.isNotBlank(bspKitQueryResult.getDeactivationDate())) {
+            logger.info("Kit can not be received.");
+            return false;
+        }
+        return true;
+
+    }
+
+    public BSPKitInfo receiveBSPKit(String kitLabel, NotificationUtil notificationUtil) {
```
gmakharat: too many things coupled to each other in one method which makes it so complex and confusing, this complexity could be cut by removing `TransactionWrapper.inTransaction`, since `BSPKit` is business logic class it is bad idea to have direct connection to database from it, instead it is better to use composition with `Dao/Dto` classes.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,150 @@
...
+                        bspMaterialType,
+                        bspReceptacleType);
+            }
+            else {
+                throw new RuntimeException("No participant id for " + kitLabel + " from " + bspKitQueryResult.getInstanceName());
+            }
+        });
+    }
+
+    private static void writeSampleReceivedToES(DDPInstance ddpInstance, BSPKitQueryResult bspKitInfo) {
```
gmakharat: access modifier static can be removed easily, since BSPKit is following OOP and this method is part of its internal business logic.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,150 @@
...
+        ElasticSearchUtil.writeSample(ddpInstance, kitRequestId, bspKitInfo.getDdpParticipantId(), ESObjectConstants.SAMPLES,
+                ESObjectConstants.KIT_REQUEST_ID, nameValuesMap);
+    }
+
+    private void triggerDDP(Connection conn, @NonNull BSPKitQueryResult bspKitInfo, boolean firstTimeReceived, String kitLabel) {
+        try {
+            if (bspKitInfo.isHasParticipantNotifications() && firstTimeReceived) {
+                KitDDPNotification kitDDPNotification = KitDDPNotification.getKitDDPNotification(TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.GET_RECEIVED_KIT_INFORMATION_FOR_NOTIFICATION_EMAIL), kitLabel, 1);
+                if (kitDDPNotification != null) {
+                    EventUtil.triggerDDP(conn, kitDDPNotification);
```
gmakharat, line 152: cutting connection here would be great as well.

src/main/java/org/broadinstitute/dsm/route/BSPKitQueryRoute.java
```diff
@@ -2,157 +2,39 @@
...
-                    else {
-                        triggerDDP(conn, bspKitInfo, firstTimeReceived, kitLabel);
-                    }
-                }
-            }
-            else {
-                triggerDDP(conn, bspKitInfo, firstTimeReceived, kitLabel);
+        BSPKit bspKit = new BSPKit();
+        if (!bspKit.canReceiveKit(kitLabel)) {
+            Optional<Object> result = Optional.ofNullable(bspKit.getKitStatus(kitLabel, notificationUtil));
```
gmakharat: Instead of wrapping result of `getKitStatus` here, as mentioned above, make method `getKitStatus` to return Optional is better.

src/main/java/org/broadinstitute/dsm/route/BSPKitQueryRoute.java
```diff
@@ -2,157 +2,39 @@
...
+            if(!result.isPresent()){
```
gmakharat: Optional's `isEmpty` method would fit more here.

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,91 @@
...
+        return getClinicalKit(kitLabel);
+    }
+
+    private ClinicalKitDto getClinicalKit(String kitLabel) {
+        logger.info("Checking label " + kitLabel);
+
+        // this method already sets the received time, check for exited and deactivation and special behaviour, and triggers DDP, we don't need a new
+        BSPKit bspKit = new BSPKit();
+        ClinicalKitDto clinicalKit = new ClinicalKitDto();
+        Optional<BSPKitInfo> maybeKitInfo = Optional.of(bspKit.receiveBSPKit(kitLabel, notificationUtil));
```
gmakharat: the same here, return Optional from `receiveBSPKit` instead of wrapping it everywhere.

src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -0,0 +1,91 @@
...
+                            clinicalKit.setDateOfBirth(dsm.get(DATE_OF_BIRtH));
+                        }
+                        Map<String, String> participantProfileFromEs = (Map<String, String>) p.getData().get(ElasticSearchUtil.PROFILE);
+                        String firstName = participantProfileFromEs.get(FIRSTNAME);
+                        String lastName = participantProfileFromEs.get(LASTNAME);
+                        String mailToName = firstName + " " + lastName;
+                        clinicalKit.setMailToName(mailToName);
+                    }
+
+                    , () -> new RuntimeException("Participant doesn't exist / is not valid for kit " + kitLabel)
```
gmakharat: I think you missed `throw` keyword here.

src/main/java/org/broadinstitute/dsm/route/CreateBSPDummyKitRoute.java
```diff
@@ -0,0 +1,61 @@
...
+            String collaboratorSampleId = KitRequestShipping.getCollaboratorSampleId(kitTypeId, participantCollaboratorId, DUMMY_KIT_TYPE_NAME);
+            if (ddpParticipantId != null) {
+                //if instance not null
+                String dsmKitRequestId = KitRequestShipping.writeRequest(mockDdpInstance.getDdpInstanceId(), mercuryKitRequestId, kitTypeId,
+                        ddpParticipantId, participantCollaboratorId, collaboratorSampleId,
+                        USER_ID, "", "", "", false, "");
+                new BSPKitDao().updateKitLabel(kitLabel, dsmKitRequestId);
+            }
+            logger.info("Returning 200 to Mercury");
+           response.status(200);
```
gmakharat: I think you need to return result from here, otherwise once if block will be passed, response status will become 500 anyway.

src/main/java/org/broadinstitute/dsm/route/CreateClinicalDummyKitRoute.java
```diff
@@ -0,0 +1,74 @@
...
+    private static String CLINICAL_KIT_PREFIX = "CLINICALKIT_";
+    private int REALM;
+    private static final String USER_ID = "MERCURY";
+
+    @Override
+    public Object handle(Request request, Response response) {
+        String kitLabel = request.params(RequestParameter.LABEL);
+        if (StringUtils.isBlank(kitLabel)) {
+            logger.warn("Got a create Clinical Kit request without a kit label!!");
+            return new Result(500, "Please include a kit label as a path parameter");
```
gmakharat: I think this should be okay:
`response.status(500);`
`return "Please include a kit label as a path parameter";`
if frontend is not depend on object of `Result` of course.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitQueryResultDao.java
```diff
@@ -0,0 +1,90 @@
...
+    public static final String BSP_PARTICIPANT_ID = "bsp_collaborator_participant_id";
+    public static final String INSTANCE_NAME = "instance_name";
+    public static final String BSP_COLLECTION = "bsp_collection";
+    public static final String BSP_ORGANISM = "bsp_organism";
+    public static final String DDP_PARTICIPANT_ID = "ddp_participant_id";
+    public static final String MATERIAL_TYPE = "bsp_material_type";
+    public static final String RECEPTACLE_TYPE = "bsp_receptacle_type";
+    public static final String PARTICIPANT_EXIT = "ddp_participant_exit_id";
+
+    public static Optional<BSPKitQueryResultDto> getBSPKitQueryResult (@NonNull String kitLabel) {
```
gmakharat: make it non static

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,157 @@
...
+            Optional<BSPKitQueryResultDto> bspKitQueryResult = BSPKitQueryResultDao.getBSPKitQueryResult(kitLabel);
+             BSPKitQueryResultDto maybeBspKitQueryResult = bspKitQueryResult.get();
+            if (StringUtils.isNotBlank(maybeBspKitQueryResult.getDdpParticipantId())) {
+                boolean firstTimeReceived = KitUtil.setKitReceived(conn, kitLabel);
+                DDPInstance ddpInstance = DDPInstance.getDDPInstance(maybeBspKitQueryResult.getInstanceName());
+                InstanceSettings instanceSettings = new InstanceSettings();
+                InstanceSettingsDto instanceSettingsDto = instanceSettings.getInstanceSettings(maybeBspKitQueryResult.getInstanceName());
+                Value received = instanceSettingsDto
+                        .getKitBehaviorChange()
+                        .map(kitBehavior -> kitBehavior.stream().filter(o -> o.getName().equals(InstanceSettings.INSTANCE_SETTING_RECEIVED)).findFirst().orElse(null))
```
gmakharat: you can remove `orElse(null)` and change `map` to `flatMap`.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,157 @@
...
+                        .orElse(null);
```
gmakharat: instead of using `orElse(null)` here, I would use `ifPresentOrElse` method which would bring much better structure than `if (received != null) {...} else {...}`.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitQueryResultDao.java
```diff
@@ -0,0 +1,90 @@
...
+            catch (Exception ex) {
+                dbVals.resultException = ex;
+            }
+            return dbVals;
+        });
+
+        if (results.resultException != null) {
+            throw new RuntimeException("Error looking up kit info for kit " + kitLabel, results.resultException);
+        }
+        return Optional.of((BSPKitQueryResultDto) results.resultValue);
```
gmakharat: `Optiona.ofNullable` is better option here, since `results.resultValue` could hold `null` value.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,157 @@
...
+                return false;
+            }
+        }
+        return true;
+
+    }
+
+    public Optional<BSPKitInfo> receiveBSPKit(String kitLabel, NotificationUtil notificationUtil) {
+        logger.info("Trying to receive kit " + kitLabel);
+        return TransactionWrapper.inTransaction(conn -> {
```
gmakharat: try to cut direct connections as much as possible, anyway such technique should be eliminated in whole project and if we do it slowly, step by step, will make refactoring process much easier.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,157 @@
...
+            Optional<BSPKitQueryResultDto> bspKitQueryResult = BSPKitQueryResultDao.getBSPKitQueryResult(kitLabel);
+             BSPKitQueryResultDto maybeBspKitQueryResult = bspKitQueryResult.get();
```
gmakharat: calling `.get()` method of `Optional` without checking if it presents, may throw `NoSuchElementException`, use either `isPresent` or `ifPresent`.

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
pegahtah: I think we would want triggerDDP to happen within the same transaction.

gmakharat: what's the point of triggering withing the same transaction?

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,156 @@
...
+            }
+        }
+        return true;
+    }
+
+    public Optional<BSPKitInfo> receiveBSPKit(String kitLabel, NotificationUtil notificationUtil) {
+        logger.info("Trying to receive kit " + kitLabel);
+        BSPKitQueryResultDao bspKitQueryResultDao = new BSPKitQueryResultDao();
+        Optional<BSPKitQueryResultDto> bspKitQueryResult = bspKitQueryResultDao.getBSPKitQueryResult(kitLabel);
+        final Optional<BSPKitInfo>[] bspKitInfo = new Optional[] {Optional.empty()};
```
gmakharat: this statement makes confusion, just because you were not able to change `bspKitInfo` value in `ifPresent` making array of it to manipulate it is a bad idea, better idea is to use `map` on `bspKitQueryResult`,  all the logic which you currently have in `ifPresent` would be placed in `map` and you would return result directly instead of putting it on zero index, if there would not be an element in `bspKitQueryResult` you would use `orElseGet`,
`bspKitQueryResult.map(bspKitQueryResult - > ... return Optional.of(new BspKitInfo(...))).orElseGet(Optional::Empty)`

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+        return true;
+    }
+
+    public Optional<BSPKitInfo> receiveBSPKit(String kitLabel, NotificationUtil notificationUtil) {
+        logger.info("Trying to receive kit " + kitLabel);
+        BSPKitQueryResultDao bspKitQueryResultDao = new BSPKitQueryResultDao();
+        Optional<BSPKitQueryResultDto> bspKitQueryResult = bspKitQueryResultDao.getBSPKitQueryResult(kitLabel);
+        bspKitQueryResult.map(maybeBspKitQueryResult -> {
+            if (StringUtils.isNotBlank(maybeBspKitQueryResult.getDdpParticipantId())) {
```
gmakharat: using guard condition instead of putting huge logic in `if` block is better for visibility:
```
if (StringUtils.isBlank(maybeBspKitQueryResult.getDdpParticipantId())) throw new RuntimeException()
//logic goes here
```

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitDao.java
```diff
@@ -0,0 +1,82 @@
+package org.broadinstitute.dsm.db.dao.kit;
```
SimoneMaiwald, line 1: I wonder if we should rename that to KitDao and lose the BSP.

SimoneMaiwald: If it is just used for the dummy kit creation maybe we should add Dummy to the name

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitQueryResultDao.java
```diff
@@ -0,0 +1,130 @@
+package org.broadinstitute.dsm.db.dao.kit;
```
SimoneMaiwald: I think we should rename **that** to BSPKitDao

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPKitQueryResultDao.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public final String SQL_UPDATE_KIT_RECEIVED = "UPDATE ddp_kit kit INNER JOIN( SELECT dsm_kit_request_id, MAX(dsm_kit_id) AS kit_id " +
+            "FROM ddp_kit GROUP BY dsm_kit_request_id) groupedKit ON kit.dsm_kit_request_id = groupedKit.dsm_kit_request_id " +
+            "AND kit.dsm_kit_id = groupedKit.kit_id SET receive_date = ?, receive_by = ? WHERE kit.receive_date IS NULL AND kit.kit_label = ?";
+
+    public Optional<BSPKitQueryResultDto> getBSPKitQueryResult(@NonNull String kitLabel) {
+        SimpleResult results = inTransaction((conn) -> {
+            SimpleResult dbVals = new SimpleResult();
+            try {
+                try (PreparedStatement stmt = conn.prepareStatement(TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.GET_BSP_RESPONSE_INFORMATION_FOR_KIT))) {
```
SimoneMaiwald: move the GET_BSP_RESPONSE_INFORMATION_FOR_KIT query to the class here and remove it from application.conf

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
SimoneMaiwald: I have to say I like the "guard condition" =)

src/main/java/org/broadinstitute/dsm/route/BSPKitQueryRoute.java
```diff
@@ -2,157 +2,41 @@
```
SimoneMaiwald: can we remove the Query in the name =)

src/main/java/org/broadinstitute/dsm/model/BSPKit.java
pegahtah: so if anything goes wrong during the whole procedure it is not marked as trigger and that can happen again. at least this is for the UPS test boston kits,  @SimoneMaiwald what do you think?

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPDummyKitDao.java
```diff
@@ -0,0 +1,99 @@
...
+                dbVals.resultException = e;
+            }
+            return dbVals;
+        });
+        if (results.resultException != null) {
+            throw new RuntimeException("Error updating kit  label for " + dsmKitRequestId, results.resultException);
+        }
+    }
+
+    public String getRandomParticipantForStudy(DDPInstance ddpInstance){
```
gmakharat, line 51: I think this method is more related to business logic but since this whole `DAO` class is marked as dummy we may keep it.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPDummyKitDao.java
```diff
@@ -0,0 +1,99 @@
...
+        String ddpParticipantId = new BSPDummyKitDao().getRandomParticipantIdForStudy(ddpInstance.getDdpInstanceId()).orElseThrow(() -> {
```
gmakharat, line 52: I don't think that you need to create `new BSPDummyKitDao()`, you are already in this class and the method itself is instance method as well.

src/main/java/org/broadinstitute/dsm/db/dao/kit/BSPDummyKitDao.java
```diff
@@ -0,0 +1,99 @@
...
+            throw new RuntimeException("Random participant id was not generated");
+        });
+        Optional<ElasticSearch> maybeParticipantByParticipantId = ElasticSearchUtil.getParticipantESDataByParticipantId(ddpInstance.getParticipantIndexES(), ddpParticipantId);
+        while (maybeParticipantByParticipantId.isEmpty() || maybeParticipantByParticipantId.get().getProfile().map(ESProfile::getHruid).isEmpty()){
+            ddpParticipantId = new BSPDummyKitDao().getRandomParticipantIdForStudy(ddpInstance.getDdpInstanceId()).orElseThrow(() -> {
```
gmakharat, line 57: the same here.

src/main/java/org/broadinstitute/dsm/db/dto/kit/BSPKitDto.java
```diff
@@ -0,0 +1,49 @@
...
+    private final String bspCollection;
+    private final String ddpParticipantId;
+    private String bspMaterialType;
+    private String bspReceptacleType;
+    private boolean hasParticipantNotifications;
+    private String participantExitId;
+    private String deactivationDate;
+    private String notificationRecipient;
+
+    public BSPKitDto(String instanceName,
```
gmakharat, line 21: I prefer to use `Builder` design pattern when it comes such complex object construction, `Builder` has many benefits including much more understanding while creating of object since it follows `withBaseUrl` standard and there is a less chance that different value will be passed to different field, except that `Builder` provides simplicity while testing, since you don't have to use constructor which has such many parameters.

gmakharat (approved):

PR #248, opened 2021-07-16
DDP-6591 | family member workflow only for shared email

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-6591

**Things have done:**

- Created generalized `AddFamilyMember` class with default implementation of adding family member and its things, it can be used as a parent class and if some studies have their own specific rules of adding family members can inherit from this class, use some of its implementation and add theirs as well.
- Created `RgpAddFamilyMember` which inherits from `AddFamilyMember` class and adds its specific elastic export rule to already existing `AddFamilyMember` functionality.
- Created method `hasFamilyMemberApplicantEmail` to determine whether family member shares email of applicant or not.
- Refactored code.


src/main/java/org/broadinstitute/dsm/model/familymember/AddFamilyMember.java
```diff
@@ -0,0 +1,169 @@
...
+    protected DDPInstanceDao ddpInstanceDao;
+    protected String studyGuid;
+    protected String participantId;
+
+    protected AddFamilyMember(AddFamilyMemberPayload addFamilyMemberPayload) {
+        this.addFamilyMemberPayload = Objects.requireNonNull(addFamilyMemberPayload);
+        this.participantData = new ParticipantData();
+        ddpInstanceDao = new DDPInstanceDao();
+        studyGuid = addFamilyMemberPayload.getRealm().orElseThrow();
+        participantId = addFamilyMemberPayload.getParticipantId().orElseThrow();
```
SimoneMaiwald: participantId or ddpParticipantId here?

src/main/java/org/broadinstitute/dsm/model/familymember/AddFamilyMember.java
```diff
@@ -0,0 +1,169 @@
...
+                    ddpInstanceDao.getEsParticipantIndexByStudyGuid(studyGuid).orElse(""),
+                    addFamilyMemberPayload.getParticipantId().orElse("")));
+        this.participantData.setData(familyMemberDetails.toMap());
+    }
+
+    protected void copyProbandData() {
+        boolean isCopyProband = addFamilyMemberPayload.getCopyProbandInfo().orElse(Boolean.FALSE);
+        if (!isCopyProband || StringUtils.isBlank(addFamilyMemberPayload.getParticipantId().orElse(""))) return;
+        Map<String, String> participantDataData = participantData.getData();
+        if (Objects.isNull(participantDataData)) throw new NoSuchElementException();
```
SimoneMaiwald, line 80: I have to say, I am not a fan of "if" without "{ }"

src/main/java/org/broadinstitute/dsm/model/familymember/AddFamilyMember.java
```diff
@@ -0,0 +1,169 @@
...
+                            new WorkflowForES.StudySpecificData(
+                                    addFamilyMemberPayload.getData().get().getCollaboratorParticipantId(),
+                                    addFamilyMemberPayload.getData().get().getFirstName(),
+                                    addFamilyMemberPayload.getData().get().getLastName()));
+            ElasticSearchUtil.writeWorkflow(instanceWithStudySpecificData, false);
+        });
+    }
+
+    protected void exportDefaultWorkflowsForFamilyMemberToES() {
+        logger.info("Exporting workflow for family member of participant: " + participantId + " to ES");
```
SimoneMaiwald: should we check here if email is equal of applicant emails?

src/main/java/org/broadinstitute/dsm/model/familymember/AddFamilyMember.java
gmakharat: And in my opinion when there is only one line it is much more readable than wrapping in curly braces, it would be perfect if we would make some kind of planning session and would agree on some code writing convention on which whole team would follow.

SimoneMaiwald: oh, I think we already said we want to do that… but then … maybe after RGP is all set we should do that

SimoneMaiwald (approved):

pegahtah (approved):

PR #250, opened 2021-07-20
DDP-6591 | family member workflow only for shared email

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-6591

**Things have done:**

- Added method `controlWorkflowByEmail` which checks if changed email matches to applicants or not and removes or adds workflows.


src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -308,6 +314,52 @@ else if (Patch.DDP_PARTICIPANT_ID.equals(patch.getParent())) {
...
+                    new org.broadinstitute.dsm.model.participant.data.ParticipantData(Integer.parseInt(patch.getId()),
+                            patch.getParentId(), Integer.parseInt(ddpInstanceByGuid.getDdpInstanceId()), patch.getFieldId(),
+                            pData);
+            if (participantData.hasFamilyMemberApplicantEmail()) {
+                int ddpInstanceIdByGuid = Integer.parseInt(ddpInstanceByGuid.getDdpInstanceId());
+                FieldSettings fieldSettings = new FieldSettings();
+                pData.forEach((columnName,columnValue) -> {
+                    if (!fieldSettings.isColumnExportable(ddpInstanceIdByGuid, columnName)) return;
+                    if (!patch.getFieldId().contains(org.broadinstitute.dsm.model.participant.data.ParticipantData.FIELD_TYPE)) return;
+                    ElasticSearchUtil.writeWorkflow(WorkflowForES.createInstanceWithStudySpecificData(ddpInstanceByGuid,
```
pegahtah, line 331: Is there. a reason for not importing the `org.broadinstitute.dsm.model.participant.data.ParticipantData` once?

pegahtah (approved):

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
gmakharat: Yes, there are two classes with the same name(ParticipantData) in different packages, PatchRoute uses both of them so to make differentiate which one is which one here is used full name of the class.

SimoneMaiwald (approved):

PR #255, opened 2021-07-22
DDP-6615 removing the 6 months limit and fixing the ES address bug

pegahtah:
removing the 6 months limit on tracking the kits so we track everything that is left

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/careevolve/Address.java
+++ b/src/main/java/org/broadinstitute/dsm/careevolve/Address.java
@@ -6,19 +6,19 @@ import org.apache.commons.lang3.StringUtils;
 
 public class Address {
 
-    @SerializedName("Line1")
+    @SerializedName ("Line1")
     private String line1;
 
-    @SerializedName("Line2")
+    @SerializedName ("Line2")
     private String line2;
 
-    @SerializedName("City")
+    @SerializedName ("City")
     private String city;
 
-    @SerializedName("State")
+    @SerializedName ("State")
     private String state;
 
-    @SerializedName("ZipCode")
+    @SerializedName ("ZipCode")
     private String zipCode;
 
     public Address(String line1, String line2, String city, String state, String zipCode) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/dsm/careevolve/Address.java
```diff
@@ -33,7 +33,7 @@ public Address(String line1, String line2, String city, String state, String zip
 
     public static Address fromElasticAddress(JsonObject esAddress) {
         return new Address(esAddress.get("street1").getAsString(),
-                esAddress.get("street2").getAsString(),
+                esAddress.has("street2") ? esAddress.get("street2").getAsString() : "",
```
pegahtah, line 36: `street2` is optional for address

src/main/java/org/broadinstitute/dsm/cf/TestBostonKitTrackerDispatcher.java
```diff
@@ -74,7 +74,6 @@ public void accept(PubsubMessage pubsubMessage, Context context) throws Exceptio
                 " ))";
         String SQL_AVOID_DELIVERED = " and (tracking_to_id is not null or tracking_return_id is not null ) and kit.test_result is null " +
                 " and ( ups_status_description is null or ups_status_description not like \"%Delivered%\") "+
-                " and  from_unixtime(created_date/1000) > NOW() - INTERVAL 180 DAY"+
```
zyme, line 77: If you run this query with and without the 180 day change, what does the difference look like?  Will anything break with a higher volume of stuff as a result?

pegahtah: The size of the result changes from ~7000 kits to ~13000 kits. We run the CF in different instances each with 1000 kits running parallel wise, and this query is also getting the kits in batches of that size. So I don't think we will face a memory issue.
I see that the execution time is around 1.5 min now and I have allocated 540 seconds so I don't think it will break by that.
However, if there are only 3 or 4 kits showing up everyday that are not ordered in CE, I am not sure it is worth checking all the 13k kits for them. especially since their ups label is expired and we will get lots of nulls and errors from UPS.
I can make a quick CF that takes the kit label, checks for the info manually ( collection time, external order number, short id) and orders it so we do ordering less manually.

zyme: How about we change this to 365 days instead of completely removing it?

zyme (approved):

PR #256, opened 2021-07-22
add check if participant id is altpid to get guid from it

gmakharat:


src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -22,6 +22,7 @@
 import org.broadinstitute.dsm.model.NameValue;
 import org.broadinstitute.dsm.model.Patch;
 import org.broadinstitute.dsm.model.Value;
+import org.broadinstitute.dsm.model.elasticsearch.ESProfile;
```
SimoneMaiwald: I don't think this is used

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -44,6 +44,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.awt.color.ICC_ColorSpace;
```
SimoneMaiwald: ?

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
gmakharat: actually it is used on line 612

SimoneMaiwald: File I am looking at only has 527 lines

yufengwng:
I think this PR is old. Looks like the code is already in `develop` now.

PR #258, opened 2021-07-23
Add functionality to remove workflows when is has no data or has the …

Irakli-Khaladze:
Added functionality to remove workflows when is has no data or has the subject which shouldn't be there, because it doesn't have ssame email as applicant. During export, getting objects from ES passing client. Also took RestHighLevelClient from writeSample and added test for the new method

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/export/WorkflowAndFamilyIdExporter.java
+++ b/src/main/java/org/broadinstitute/dsm/export/WorkflowAndFamilyIdExporter.java
@@ -122,20 +122,20 @@ public class WorkflowAndFamilyIdExporter implements Exporter {
             String collaboratorParticipantId = dataMap.get(FamilyMemberConstants.COLLABORATOR_PARTICIPANT_ID);
             if (!ParticipantUtil.matchesApplicantEmail(collaboratorParticipantId, participantDataFamily)) {
                 ElasticSearchUtil.removeWorkflowIfNoDataOrWrongSubject(client, ddpParticipantId, ddpInstance, collaboratorParticipantId);
-                return;
+            } else {
+                //is matching applicant email => write into ES
+                WorkflowForES.StudySpecificData studySpecificData = new WorkflowForES.StudySpecificData(
+                        collaboratorParticipantId,
+                        dataMap.get(FamilyMemberConstants.FIRSTNAME),
+                        dataMap.get(FamilyMemberConstants.LASTNAME)
+                );
+                dataMap.entrySet().stream().filter(entry -> workFlowColumnNames.contains(entry.getKey()))
+                        .forEach(entry -> {
+                            ElasticSearchUtil.writeWorkflow(client, WorkflowForES.createInstanceWithStudySpecificData(ddpInstance, ddpParticipantId,
+                                    entry.getKey(), entry.getValue(), studySpecificData), clearBeforeUpdate.get());
+                            clearBeforeUpdate.set(false);
+                        });
             }
-            //is matching applicant email => write into ES
-            WorkflowForES.StudySpecificData studySpecificData = new WorkflowForES.StudySpecificData(
-                    collaboratorParticipantId,
-                    dataMap.get(FamilyMemberConstants.FIRSTNAME),
-                    dataMap.get(FamilyMemberConstants.LASTNAME)
-            );
-            dataMap.entrySet().stream().filter(entry -> workFlowColumnNames.contains(entry.getKey()))
-                    .forEach(entry -> {
-                        ElasticSearchUtil.writeWorkflow(client, WorkflowForES.createInstanceWithStudySpecificData(ddpInstance, ddpParticipantId,
-                                entry.getKey(), entry.getValue(), studySpecificData), clearBeforeUpdate.get());
-                        clearBeforeUpdate.set(false);
-                    });
         } else {
             dataMap.entrySet().stream().filter(entry -> workFlowColumnNames.contains(entry.getKey()))
                     .forEach(entry -> {
--- a/src/main/java/org/broadinstitute/dsm/export/WorkflowAndFamilyIdExporter.java
+++ b/src/main/java/org/broadinstitute/dsm/export/WorkflowAndFamilyIdExporter.java
@@ -119,14 +119,14 @@ public class WorkflowAndFamilyIdExporter implements Exporter {
                                        String ddpParticipantId, List<ParticipantDataDto> participantDataFamily, Map<String, String> dataMap,
                                        AtomicBoolean clearBeforeUpdate) {
         if (participantData.getFieldTypeId().orElse("").equals(RGP_PARTICIPANTS)) {
-            if (!ParticipantUtil.checkApplicantEmail(dataMap.get(FamilyMemberConstants.COLLABORATOR_PARTICIPANT_ID), participantDataFamily)) {
-                //TODO doesn't match applicant Email  => remove from ES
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -672,23 +707,39 @@ public static void createAndAddNewObjectMap(@NonNull Object id, List<Map<String,
         objectList.add(newObjectMap);
     }
 
+    public static Map<String, Object> getObjectsMapHavingClient(RestHighLevelClient client, String index, String id, String object) throws Exception {
```
zyme: What is `object` here?  Can you rename with more informative variable?  Similar comment for L722.

zyme: How about `getObjectsMap` instead of the `getObjectsMapHavingClient`?

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -382,6 +382,33 @@ public static void writeWorkflow(@NonNull WorkflowForES workflowForES, boolean c
...
+        String index = ddpInstance.getParticipantIndexES();
+        try {
+            Map<String, Object> workflowMapES = getObjectsMapHavingClient(client, index, ddpParticipantId, ESObjectConstants.WORKFLOWS);
+            if (workflowMapES == null || workflowMapES.isEmpty()) {
+                return;
+            }
+            List<Map<String, Object>> workflowListES = (List<Map<String, Object>>) workflowMapES.get(ESObjectConstants.WORKFLOWS);
+            int initialListSize = workflowListES.size();
+            workflowListES.removeIf(workflow -> !workflow.containsKey("data") ||
+                    (((Map<String, Object>) workflow.get("data")).get("subjectId") != null
```
zyme: Can you change the casts to just `(Map)`? instead of `Map<String,Object>`?

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -382,6 +382,33 @@ public static void writeWorkflow(@NonNull WorkflowForES workflowForES, boolean c
...
+                            && ((Map<String, Object>) workflow.get("data")).get("subjectId").equals(collaboratorParticipantId)));
```
zyme: Can you yoda-ify this to avoid an NPE so it's

`collaboratorParticipantId.equals(((Map)workflow.get("data")).get("subjectId"))` ?

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -382,6 +382,33 @@ public static void writeWorkflow(@NonNull WorkflowForES workflowForES, boolean c
...
+    public static void removeWorkflowIfNoDataOrWrongSubject(RestHighLevelClient client, String ddpParticipantId, DDPInstance ddpInstance, String collaboratorParticipantId) {
+        String index = ddpInstance.getParticipantIndexES();
+        try {
+            Map<String, Object> workflowMapES = getObjectsMapHavingClient(client, index, ddpParticipantId, ESObjectConstants.WORKFLOWS);
+            if (workflowMapES == null || workflowMapES.isEmpty()) {
+                return;
+            }
+            List<Map<String, Object>> workflowListES = (List<Map<String, Object>>) workflowMapES.get(ESObjectConstants.WORKFLOWS);
+            int initialListSize = workflowListES.size();
+            workflowListES.removeIf(workflow -> !workflow.containsKey("data") ||
```
zyme: Please pull out strings to constants

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -382,6 +382,33 @@ public static void writeWorkflow(@NonNull WorkflowForES workflowForES, boolean c
...
+                    (((Map<String, Object>) workflow.get("data")).get("subjectId") != null
+                            && ((Map<String, Object>) workflow.get("data")).get("subjectId").equals(collaboratorParticipantId)));
+            if (workflowListES.size() == initialListSize) {
```
zyme: `removeIf()` returns a boolean indicating whether or not something was removed--can you use this instead of checking the size?

src/main/java/org/broadinstitute/dsm/export/WorkflowAndFamilyIdExporter.java
```diff
@@ -119,14 +119,14 @@ public void exportWorkflows(RestHighLevelClient client, List<String> workFlowCol
...
         if (participantData.getFieldTypeId().orElse("").equals(RGP_PARTICIPANTS)) {
-            if (!ParticipantUtil.checkApplicantEmail(dataMap.get(FamilyMemberConstants.COLLABORATOR_PARTICIPANT_ID), participantDataFamily)) {
-                //TODO doesn't match applicant Email  => remove from ES
-                //TODO workflow object doesn't have a data object => remove from ES
+            String collaboratorParticipantId = dataMap.get(FamilyMemberConstants.COLLABORATOR_PARTICIPANT_ID);
+            if (!ParticipantUtil.matchesApplicantEmail(collaboratorParticipantId, participantDataFamily)) {
+                ElasticSearchUtil.removeWorkflowIfNoDataOrWrongSubject(client, ddpParticipantId, ddpInstance, collaboratorParticipantId);
                 return;
             }
             //is matching applicant email => write into ES
```
zyme: How about putting an `else` here instead of just relying on early return above?

src/test/java/org/broadinstitute/dsm/ElasticSearchTest.java
```diff
@@ -787,6 +789,32 @@ public void createTestParticipantsInES() throws Exception {
         }
     }
 
+    @Test
+    public void testRemoveWorkflowIfNoDataOrWrongSubject() throws Exception {
```
zyme, line 793: Nice tests!

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -672,23 +707,39 @@ public static void createAndAddNewObjectMap(@NonNull Object id, List<Map<String,
...
+                .index(index)
+                .type("_doc")
+                .id(id)
+                .fetchSourceContext(fetchSourceContext);
+
+        GetResponse getResponse = null;
+        if (client.exists(getRequest, RequestOptions.DEFAULT)) {
+            getResponse = client.get(getRequest, RequestOptions.DEFAULT);
+        }
+        return getResponse != null ? getResponse.getSourceAsMap() : null;
```
zyme: Might avoid client NPEs by returning `Collections.emptyMap()` instead of `null`.

zyme (commented):
Overall looks good functionality-wise, just some cosmetics, code clarity, and suggestions for avoiding potential NPEs down the road.

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
Irakli-Khaladze: I guess I don't need getObjectsMapHavingClient at all? Maybe I delete it?

zyme: If it's not needed by anything, sure!

PR #259, opened 2021-07-23
remove applicant field

gmakharat:


src/main/java/org/broadinstitute/dsm/model/rgp/AutomaticProbandDataCreator.java
```diff
@@ -116,7 +116,6 @@ private boolean extractAndInsertProbandFromESData(DDPInstance instance, ElasticS
                         new FamilyMemberDetails(firstName, lastName, memberType, familyId, collaboratorParticipantId);
                 probandMemberDetails.setMobilePhone(mobilePhone);
                 probandMemberDetails.setEmail(email);
-                probandMemberDetails.setApplicant(true);
```
yufengwng, line 119: I think we're using `is_applicant` in `ParticipantUtil.matchesApplicantEmail()`?

PR #260, opened 2021-07-24
DDP-6643 fix ES lookup call when removing workflows in patch-route

yufengwng:
Fixes DDP-6643.

When the email doesn't match the applicant email, we make a call to ES to fetch the workflow. The bug is that we're using the altpid for the doc ID instead of the participant guid. And in this case, for that particular account, we happened to have an old ES document with the altpid as doc ID. In that old document, we have workflows for one of the family members (who is not the one we're patching). Given how we're iterating and remove things from the workflow list, we ended up having an empty list. The sizes doesn't match up, so we issue an update request, and now we have no workflows.

This PR will fix the bug in the code, but we should also delete and recreate the indices for RGP in prod to get rid of any old/stale documents.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
+++ b/src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
@@ -22,6 +22,7 @@ import org.broadinstitute.dsm.model.AbstractionWrapper;
 import org.broadinstitute.dsm.model.NameValue;
 import org.broadinstitute.dsm.model.Patch;
 import org.broadinstitute.dsm.model.Value;
+import org.broadinstitute.dsm.model.elasticsearch.ESProfile;
 import org.broadinstitute.dsm.model.settings.field.FieldSettings;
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -335,8 +343,9 @@ private void controlWorkflowByEmail(Patch patch, NameValue nameValue) {
                                     pData.get(FamilyMemberConstants.LASTNAME))), false);
                 });
             } else {
+                logger.info("Email in patch data does not match participant profile email, will remove workflows");
                 Map<String, Object> esMap = ElasticSearchUtil
-                        .getObjectsMap(ddpInstanceByGuid.getParticipantIndexES(), patch.getParentId(),
+                        .getObjectsMap(ddpInstanceByGuid.getParticipantIndexES(), profile.getParticipantGuid(),
```
yufengwng, line 348: This is the bug.

SimoneMaiwald (approved):

PR #261, opened 2021-07-24
refactor: reuse elasticsearch client

yufengwng:
Let's minimize creation of elasticsearch client and reuse a single instance as much as possible. It seems fine when I run this locally, but will have to do more testing on `dev`.

CC @pegahtah @gmakharat @Irakli-Khaladze

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
@@ -97,6 +97,28 @@ public class ElasticSearchUtil {
     public static final String WORKFLOWS = "workflows";
     public static final String EMAIL_FIELD = "email";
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -97,6 +97,28 @@
     public static final String WORKFLOWS = "workflows";
     public static final String EMAIL_FIELD = "email";
 
+    // These clients are expensive. They internally have thread pools and other resources. Let's
+    // create one instance and reuse it as much as possible. Client is thread-safe per the docs.
```
yufengwng, line 101: The docs: https://www.elastic.co/guide/en/elasticsearch/client/java-rest/6.8/_changing_the_client_8217_s_initialization_code.html

SimoneMaiwald (approved):

PR #262, opened 2021-07-24
refactor: optimize workflows export

yufengwng:
I noticed that we're have inefficient looping and doing a ton of ES calls. This PR hopefully will make it better.

CC @pegahtah @gmakharat @Irakli-Khaladze

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/export/WorkflowsEditor.java
+++ b/src/main/java/org/broadinstitute/dsm/export/WorkflowsEditor.java
@@ -1,0 +1,73 @@
+package org.broadinstitute.dsm.export;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.broadinstitute.dsm.statics.ESObjectConstants;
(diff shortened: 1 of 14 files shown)


src/main/java/org/broadinstitute/dsm/export/WorkflowsEditor.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.broadinstitute.dsm.statics.ESObjectConstants;
+import org.broadinstitute.dsm.util.ElasticSearchUtil;
+
+/**
+ * A container that provides a nicer interface to updating the workflows list. The interface is similar to the "builder" pattern.
+ */
+public class WorkflowsEditor {
```
yufengwng, line 14: Right now, editing the workflows list is still not ideal because we're still iterating the list a ton of times. But hopefully it should be okay for now since the workflows list tends to be small enough. One way we can probably make this better is to invert our data structures so that instead of a list of maps, we have a mapping of `subject_id -> map<workflow, status>`.

SimoneMaiwald (approved):

PR #264, opened 2021-07-24
refactor: fix workflows editor in export

yufengwng:
Fixed small issue in export.


src/main/java/org/broadinstitute/dsm/export/WorkflowsEditor.java
```diff
@@ -37,7 +41,8 @@ public WorkflowsEditor upsert(WorkflowForES workflow) {
             updated = ElasticSearchUtil.updateWorkflow(
                     workflow.getWorkflow(), workflow.getStatus(), workflows);
         }
-        hasChanged = hasChanged || updated;
+        boolean added = workflows.size() > oldSize;
+        hasChanged = hasChanged || updated || added;
```
yufengwng, line 45: This was the bug. I didn't handle the "insert" case of upsert.

PR #268, opened 2021-07-26
More than 1000 guids bugfix

Irakli-Khaladze:
Logic explanation:
Previously for each of the filter related to dynamic forms we were getting separate AND group of guids like this: AND (guid1 or guid2) AND (guid3 or guid1 or guid 2) AND ... Added hashmap to use the fact that  AND (guid1 or guid2) AND (guid3 or guid1 or guid 2) = AND (guid1 or guid2). So currently for all the filter related to dynamic filds we are getting only one AND group with guids separated by OR inside it. So implemented changes in addParticipantDataFilters and introduced a new method as well.

After that in ParticipantWrapper we needed to divide this AND group if it had too many guid, but we might also have other AND groups not related to dynamic fields, so I'm using the fact that AND (bool1) AND (bool2) AND (bool3 or bool4) = ((bool1) AND (bool2) AND (bool3)) OR ((bool1) AND (bool2) AND (bool4)). So we are dividing the large AND group of participantIds and query one by one with other conditions and add the participants. After that send back those participants.

Filter change alert: will need to be tested again after it's reviewed, corrected and merged

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
+++ b/src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
@@ -172,9 +172,9 @@ public class ParticipantWrapper {
                 //get only pts for the filtered data
                 if (baseList != null && !baseList.isEmpty()) {
                     //ES can only filter for 1024 (too_many_clauses: maxClauseCount is set to 1024)
-                    if (baseList.size() > 1000) {
+                    if (baseList.size() > Filter.THOUSAND) {
                         //make sub-searches
-                        Collection<List<String>> partitionBaseList = partitionBasedOnSize(baseList, 1000);
+                        Collection<List<String>> partitionBaseList = partitionBasedOnSize(baseList, Filter.THOUSAND);
                         for (Iterator i = partitionBaseList.iterator(); i.hasNext();) {
                             List<String> baseListPart = ((List<String>) i.next());
                             participantESData = addParticipantESData(instance, baseListPart, participantESData, ElasticSearchUtil.BY_GUID, ElasticSearchUtil.BY_GUIDS);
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -161,7 +161,31 @@ else if (DBConstants.DDP_ABSTRACTION_ALIAS.equals(source)) {
...
+                                        .getFilteredDDPParticipantsFromES(instance, lessThanLimit + " AND " + String.join(Filter.OR, baseListPart));
+                                if (tempParticipantESData != null) {
+                                    participantESData.putAll(tempParticipantESData);
+                                }
+                            }
+                            break;
+                        }
+                        if (!tooManyParameters) {
+                            participantESData = ElasticSearchUtil.getFilteredDDPParticipantsFromES(instance, filters.get(source));
+                        }
```
gmakharat: extracting whole logic into method with meaningful name would be good.

src/main/java/org/broadinstitute/dsm/route/FilterRoute.java
```diff
@@ -296,8 +300,32 @@ public Object processRequest(Request request, Response response, String userId)
...
+                    newCondition.append(ParticipantUtil.isGuid(entry.getKey()) ? ElasticSearchUtil.BY_PROFILE_GUID + entry.getKey() : ElasticSearchUtil.BY_PROFILE_LEGACY_ALTPID + entry.getKey());
+                } else {
+                    newCondition.append(ParticipantUtil.isGuid(entry.getKey()) ? ElasticSearchUtil.BY_GUIDS + entry.getKey() : ElasticSearchUtil.BY_LEGACY_ALTPIDS + entry.getKey());
+                }
+                i++;
+            }
+            newCondition.append(ElasticSearchUtil.CLOSING_PARENTHESIS);
+            String esCondition = queryConditions.get(ElasticSearchUtil.ES);
+            esCondition += newCondition.toString();
+            queryConditions.put(ElasticSearchUtil.ES, esCondition);
```
gmakharat: code from line 321 to 324 could be minimized using Map's `computeIfPresent`,
`queryConditions.computeIfPresent(ElasticSearchUtil.ES, (key, value) -> value += newCondition.toString())`,
except that `else` block's content could be extracted as a method with name something like `createParticipantIdsQuery` or something like that, also `i` and `newCondition` is used only in `else` block so they can be easily moved to extracted method.

gmakharat (approved):

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -246,6 +246,36 @@ else if (DBConstants.DDP_ABSTRACTION_ALIAS.equals(source)) {
...
+        String[] betweenAnds = wholeFilter.split(Filter.AND_TRIMMED);
+        boolean tooManyParameters = false;
+        participantESData = new HashMap<>();
+        String lessThanLimit = Arrays.stream(betweenAnds).filter(query -> query.split(Filter.OR_TRIMMED).length <= 900).collect(Collectors.joining(Filter.AND));
+        for (String query: betweenAnds) {
+            String[] orQueries = query.split(Filter.OR_TRIMMED);
+            if (orQueries.length <= 900) {
+                continue;
+            }
+            tooManyParameters = true;
```
SimoneMaiwald, line 261: looks like it will be always true

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
```diff
@@ -246,6 +246,36 @@ else if (DBConstants.DDP_ABSTRACTION_ALIAS.equals(source)) {
         }
     }
 
+    public static Map<String, Map<String, Object>> getParticipantESDataConsideringNumberOfParameters(@NonNull DDPInstance instance, Map<String, String> filters, String source) {
+        Map<String, Map<String, Object>> participantESData;
+        String wholeFilter = filters.get(source);
+        String[] betweenAnds = wholeFilter.split(Filter.AND_TRIMMED);
+        boolean tooManyParameters = false;
+        participantESData = new HashMap<>();
+        String lessThanLimit = Arrays.stream(betweenAnds).filter(query -> query.split(Filter.OR_TRIMMED).length <= 900).collect(Collectors.joining(Filter.AND));
```
SimoneMaiwald: ES query max is 1024 do we want to change the 900 to 1000? maybe make a parameter which can be re-used?

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
Irakli-Khaladze: It will be false if none of the between AND-s groups have too many parameters

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
Irakli-Khaladze: okay, I'll change it to 1000

src/main/java/org/broadinstitute/dsm/model/ParticipantWrapper.java
SimoneMaiwald: :doh: I missed the continue in the if
Sorry!

SimoneMaiwald (approved):

PR #269, opened 2021-07-27
DDP-6662 disable testboston ce ordering

zyme:
DDP-6662 calls for disabling CE ordering for longitudinal testboston kits after a certain date TBD.  The implementation here is to read that value out of the config value, and if we are beyond the date, we then check to see whether the kit has an `uploadReason`, which is what we've been using to differentiate scheduled orders from adhoc orders in TestBoston.


src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -508,6 +543,43 @@ private void updateTrackingInfo(PoolingDataSource<PoolableConnection> dataSource
 
     }
 
+    /**
+     * Queries the database to see whether there is an upload reason
+     * for the given kit label
+     */
+    private boolean doesKitHaveUploadReason(Connection conn, String kitLabel) throws SQLException {
+        boolean hasUploadReason = false;
+        try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_KIT_REQUEST + BY_DDP_LABEL)) {
```
zyme: I decided to use existing queries for this--if there's risk that such a query might pick up non-tbos orders, please let me know and we can go by `dsmKitRequestId` or something else.

src/main/java/org/broadinstitute/dsm/db/dto/ddp/kitrequest/KitRequestDto.java
```diff
@@ -59,4 +60,8 @@ public KitRequestDto(
         this.uploadReason = uploadReason;
         this.orderTransmitted_at = orderTransmitted_at;
     }
+
+    public boolean hasUploadReason() {
+        return StringUtils.isBlank(uploadReason);
```
yufengwng: Don't we want the negation here?
```suggestion
        return StringUtils.isNotBlank(uploadReason);
```

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
pegahtah: I think instead of this query, you want to use `SELECT * FROM FROM ddp_kit_request req LEFT JOIN ddp_kit kit on (kit.dsm_kit_request_id = req.dsm_kit_request_id) WHERE kit.kit_label = ?`
the `ddp_label` is null in test boston

src/main/java/org/broadinstitute/dsm/model/birch/RegisterBirchOrderCLI.java
```diff
@@ -0,0 +1,138 @@
...
+            String collectionTime = args[3];
+
+            Instant collectionDate = null;
+            try {
+                collectionDate = new SimpleDateFormat(DATE_FORMAT).parse(collectionTime).toInstant();
+            } catch(ParseException e) {
+                throw new RuntimeException("Could not parse collection time '"  + collectionTime + "'.  Please use " + DATE_FORMAT + ".  You may need to quote the whole time.");
+            }
+
+            Config cfg = ConfigFactory.load();
```
pegahtah: I think  here we need to load vault config file as well, such as `cfg = cfg.withFallback(ConfigFactory.parseFile(new File("appengine/deploy/vault.conf")));` to access `portal.dbUrl`

zyme: Eeek not sure how I let this one into branch--removing this unrelated file.

src/main/java/org/broadinstitute/dsm/db/dto/ddp/kitrequest/KitRequestDto.java
```diff
@@ -59,4 +60,8 @@ public KitRequestDto(
         this.uploadReason = uploadReason;
         this.orderTransmitted_at = orderTransmitted_at;
     }
+
+    public boolean hasUploadReason() {
+        return StringUtils.isNotBlank(uploadReason);
```
zyme, line 65: Thanks @yufengwng !

src/main/java/org/broadinstitute/dsm/jobs/TestBostonUPSTrackingJob.java
```diff
@@ -508,6 +543,43 @@ private void updateTrackingInfo(PoolingDataSource<PoolableConnection> dataSource
 
     }
 
+    /**
+     * Queries the database to see whether there is an upload reason
+     * for the given kit label
+     */
+    private boolean doesKitHaveUploadReason(Connection conn, String kitLabel) throws SQLException {
+        boolean hasUploadReason = false;
+        try (PreparedStatement stmt = conn.prepareStatement(SQL_GET_KIT_REQUEST + BY_KIT_LABEL)) {
```
pegahtah, line 552: I don't think this query will work since that is only selecting from `ddp_kit_request` while `ddp_kit` has the `kit_label`. I'm afraid you have to add a query of their left join on `dsm_kit_request_id`

zyme: Oops!  Ok, will see if there's a different reusable snippet of SQL here and if not, will roll one as you suggest.

pegahtah (approved):

PR #273, opened 2021-08-03
DDP-6664: NPE 

SimoneMaiwald:
fixed npe if ES profile is empty because of wrong id used to search for it
backend part for: https://github.com/broadinstitute/ddp-study-manager-ui/pull/108

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/route/LookupRoute.java
+++ b/src/main/java/org/broadinstitute/dsm/route/LookupRoute.java
@@ -38,8 +38,7 @@ public class LookupRoute extends RequestHandler {
             "LEFT JOIN ddp_instance_group gr on (realm.ddp_instance_id = gr.ddp_instance_id) " +
             "WHERE oD.type_px LIKE ? AND NOT (oD.deleted <=> 1) AND gr.ddp_group_id = ?";
     private static final String SQL_SELECT_CONTACT = "SELECT DISTINCT name, contact, phone, fax FROM ddp_medical_record WHERE name LIKE ?";
-    private static final String SQL_SELECT_FACILITY = "SELECT DISTINCT facility, phone, fax, destruction_policy FROM ddp_onc_history_detail WHERE facility LIKE ? AND NOT (deleted <=> 1)";
-    private static final String SQL_SELECT_FACILITY_IN_GROUP = "SELECT DISTINCT oD.facility, oD.phone, oD.fax, oD.destruction_policy, realm.ddp_instance_id " +
+    private static final String SQL_SELECT_FACILITY_IN_GROUP = "SELECT DISTINCT oD.facility, oD.phone, oD.fax, oD.destruction_policy " +
             "FROM ddp_onc_history_detail oD " +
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/dsm/route/LookupRoute.java
```diff
@@ -38,8 +38,7 @@
             "LEFT JOIN ddp_instance_group gr on (realm.ddp_instance_id = gr.ddp_instance_id) " +
             "WHERE oD.type_px LIKE ? AND NOT (oD.deleted <=> 1) AND gr.ddp_group_id = ?";
     private static final String SQL_SELECT_CONTACT = "SELECT DISTINCT name, contact, phone, fax FROM ddp_medical_record WHERE name LIKE ?";
-    private static final String SQL_SELECT_FACILITY = "SELECT DISTINCT facility, phone, fax, destruction_policy FROM ddp_onc_history_detail WHERE facility LIKE ? AND NOT (deleted <=> 1)";
```
SimoneMaiwald, line 41: removed unused query

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -76,11 +76,14 @@ public Object processRequest(Request request, Response response, String userId)
                 Patch patch = gson.fromJson(requestBody, Patch.class);
                 if (StringUtils.isNotBlank(patch.getId())) {
                     //multiple values are changing
+                    DDPInstance ddpInstance = DDPInstance.getDDPInstance(patch.getRealm());
```
SimoneMaiwald, line 79: moved ddpInstance out of writeDSMRecordsToES, to not get same info twice (was L81 and in method again)

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -76,11 +76,14 @@ public Object processRequest(Request request, Response response, String userId)
...
                     if (patch.getNameValues() != null && !patch.getNameValues().isEmpty()) {
                         List<NameValue> nameValues = new ArrayList<>();
-                        DDPInstance ddpInstance = DDPInstance.getDDPInstance(patch.getRealm());
-                        ESProfile profile = ElasticSearchUtil.getParticipantProfileByGuidOrAltPid(ddpInstance.getParticipantIndexES(), patch.getParentId())
-                                .orElseThrow(() -> new RuntimeException("Unable to find ES profile for participant: " + patch.getParentId()));
+                        ESProfile profile = ElasticSearchUtil.getParticipantProfileByGuidOrAltPid(ddpInstance.getParticipantIndexES(), patch.getDdpParticipantId())
+                                .orElse(null);
+                        if (profile == null) {
+                            logger.error("Unable to find ES profile for participant with guid/altpid: {}, continuing w/ patch", patch.getParentId());
```
SimoneMaiwald, line 85: removed the throw and logging error now, because we still want to patch the value in the db even if we do not write into ES

pegahtah (approved):

PR #274, opened 2021-08-05
DDP-6701: added type radio to get default values and renaming methods…

SimoneMaiwald:
… and parameters to match the query

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/db/dao/settings/FieldSettingsDao.java
+++ b/src/main/java/org/broadinstitute/dsm/db/dao/settings/FieldSettingsDao.java
@@ -19,7 +19,7 @@ public class FieldSettingsDao implements Dao<FieldSettingsDto> {
 
     private static FieldSettingsDao fieldSettingsDao;
 
-    private static final String SQL_SETTINGS_WITH_DEFAULT_VALUE_BY_INSTANCE_ID = "SELECT " +
+    private static final String SQL_OPTIONS_AND_RADIOS_BY_INSTANCE_ID = "SELECT " +
             "field_settings_id," +
             "ddp_instance_id," +
             "field_type," +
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/dsm/db/dao/settings/FieldSettingsDao.java
```diff
@@ -19,7 +19,7 @@
 
     private static FieldSettingsDao fieldSettingsDao;
 
-    private static final String SQL_OPTIONS_BY_INSTANCE_ID = "SELECT " +
+    private static final String SQL_SETTINGS_WITH_DEFAULT_VALUE_BY_INSTANCE_ID = "SELECT " +
```
gmakharat: name looks kind of confusing, I can't see any part in SQL which filters default values.
Except that, O from SOLID principles which stand for `Open-closed principle`, says that classes should be open for extension but closed for modification, current change more or less breaks this principle since already existing behavior has been modified, maybe this current SQL and its related method were used only once, but in my opinion, instead of changing it, adding new SQL and its related method would be more convenient, we would have both behavior, already existing one which takes data by `OPTIONS` and new one which would take data by `OPTIONS or RADIO`.

SimoneMaiwald: but do we need data by `OPTIONS` ? I normally like to remove unused methods and the default values should also get set for options and radios I don't think the method to get only fieldsettings of type option is needed anymore

gmakharat: Are you sure that there won't be a case when fetching field settings by `OPTIONS` will be necessary? if so, removing it is okay. I always try to keep **existing** behavior if I am not sure that this behavior won't be used anymore, I think this way does not violate `YAGNI` principle, since it is about already **existing** behavior.

PR #275, opened 2021-08-06
hotfix for filter exception about AND ()

Irakli-Khaladze:
We are checking participant guid or altpid only in case it was added to our map as many times as there are filters related to participantData. Previously if we had participants in map, which were added less times than number of participantData related filters we would get AND (), but now we will get AND (profile.guid = empty) as it was planned before.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/model/filter/participant/BaseFilterParticipantList.java
+++ b/src/main/java/org/broadinstitute/dsm/model/filter/participant/BaseFilterParticipantList.java
@@ -114,12 +114,8 @@ public abstract class BaseFilterParticipantList extends BaseFilter implements Fi
     }
 
     public void addParticipantDataConditionsToQuery(Map<String, Integer> allIdsForParticipantDataFiltering, Map<String, String> queryConditions, int filtersLength) {
-        if (allIdsForParticipantDataFiltering.isEmpty()) {
-            queryConditions.put(ElasticSearchUtil.ES, ElasticSearchUtil.BY_PROFILE_GUID + ElasticSearchUtil.EMPTY);
-        } else {
-            String newCondition = createNewConditionByIds(allIdsForParticipantDataFiltering, filtersLength);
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsm/model/filter/participant/BaseFilterParticipantList.java
```diff
@@ -114,12 +114,8 @@ public BaseFilterParticipantList(String jsonBody) {
...
 
     public void addParticipantDataConditionsToQuery(Map<String, Integer> allIdsForParticipantDataFiltering, Map<String, String> queryConditions, int filtersLength) {
-        if (allIdsForParticipantDataFiltering.isEmpty()) {
-            queryConditions.put(ElasticSearchUtil.ES, ElasticSearchUtil.BY_PROFILE_GUID + ElasticSearchUtil.EMPTY);
-        } else {
-            String newCondition = createNewConditionByIds(allIdsForParticipantDataFiltering, filtersLength);
-            queryConditions.merge(ElasticSearchUtil.ES, newCondition, (prev, next) -> prev + next);
-        }
+        String newCondition = createNewConditionByIds(allIdsForParticipantDataFiltering, filtersLength);
+        queryConditions.merge(ElasticSearchUtil.ES, newCondition, (prev, next) -> prev + next);
```
gmakharat, line 118: concatenating previous and current value without some separator is okay? suffix of `newCondition` has separator already or maybe it does not need to be separated at all?

Irakli-Khaladze: " AND " is separator which is prefix of each new condition

gmakharat (approved):

SimoneMaiwald (approved):
Could we also make a PR for this change into develop. so that we do not need to wait till the refactoring is going to prod for it to be fixed
please and thank you

PR #278, opened 2021-08-13
DDP-6399 | improve loading performance

gmakharat:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-6399

**Things have done:**

- Added `ElasticSearchable` interface and `ElasticSearch` class which implements it, having `ElasticSearchable` interface helps to follow dependency inversion principle and to write unit tests more easily.
- Refactor `ParticipantWrapper` class, introduced `ParticipantWrapperPayload` class to wrap necessary information which is processed from frontend request, introduced `ParticipantWrapperResult` class which holds list of `ParticipantWrapperDto`(class to hold all necessary data regarding to participant to return as a response to request) and `totalCount`(which is size of participants according to filtering).
- Applied `Builder` pattern to some existing classes.
- Used `ElasticSearch`s `from`, `size` properties to provide pagination.
- Used `ElasticSearch`'s `termsQuery` to avoid `max_clause_count` exception.

P.S `ParticipantWrapper` class is not perfect yet, in the future it can be refactored to provide low coupling since it is still depend on many concrete classes.


src/main/java/org/broadinstitute/dsm/model/elasticsearch/ElasticSearch.java
```diff
@@ -1,161 +1,253 @@
...
+        return Optional.of(elasticSearchParticipantDto);
+    }
 
-        public Builder withMedicalProviders(List<Object> medicalProviders) {
-            this.medicalProviders = Optional.ofNullable(medicalProviders);
-            return this;
+    public List<ElasticSearchParticipantDto> parseSourceMaps(SearchHit[] searchHits) {
+        if (Objects.isNull(searchHits)) return Collections.emptyList();
+        List<ElasticSearchParticipantDto> result = new ArrayList<>();
+        String ddp = getDdpFromSearchHit(Arrays.stream(searchHits).findFirst().orElse(null));
```
SimoneMaiwald, line 69: that is the ddp string we display at the participant list, yes? I wonder if we should display the "display_name" from the ddp_instance table for that

gmakharat: Yes, I don't know if it worth to make `ElasticSearch` class depend on some DB related classes because of ddp name?

src/main/java/org/broadinstitute/dsm/model/filter/participant/EmptyFilterParticipantList.java
```diff
@@ -2,20 +2,34 @@
```
SimoneMaiwald, line 2: can you remove unused imports please

src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -0,0 +1,259 @@
...
+            medicalRecords = MedicalRecord.getMedicalRecordsByParticipantIds(ddpInstance.getName(), participantIds);
+            oncHistoryDetails = OncHistoryDetail.getOncHistoryDetailsByParticipantIds(ddpInstance.getName(), participantIds);
+        }
+        if (DDPInstanceDao.getRole(ddpInstance.getName(), DBConstants.KIT_REQUEST_ACTIVATED)) {
+            kitRequests = KitRequestShipping.getKitRequestsByParticipantIds(ddpInstance, participantIds);
+        }
+        abstractionActivities =
+                AbstractionActivity.getAllAbstractionActivityByParticipantIds(ddpInstance.getName(), participantIds);
+        abstractionSummary = AbstractionFinal.getAbstractionFinalByParticipantIds(ddpInstance.getName(), participantIds);
+        proxiesByParticipantIds = getProxiesWithParticipantIdsFromElasticList(ddpInstance.getParticipantIndexES(), esData.getEsParticipants());
```
SimoneMaiwald: I think this is the wrong index here. should be `usersIndexES`

src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -0,0 +1,259 @@
...
+            }
+        }
+        if (abstractionActivities.isEmpty()) {
+            abstractionActivities = AbstractionActivity.getAllAbstractionActivityByRealm(ddpInstance.getName());
+        }
+        if (abstractionSummary.isEmpty()) {
+            abstractionSummary = AbstractionFinal.getAbstractionFinal(ddpInstance.getName());
+        }
+        if (proxiesByParticipantIds.isEmpty()) {
+            proxiesByParticipantIds = getProxiesWithParticipantIdsFromElasticList(ddpInstance.getParticipantIndexES(), esData.getEsParticipants());
```
SimoneMaiwald: index should be `usersIndexES`

src/main/java/org/broadinstitute/dsm/model/elasticsearch/ElasticSearch.java
SimoneMaiwald: no, I think for now we leave it as it is. Maybe down the road we can add something to have the nice display_name, but also not in the ElasticSearch class, I think more about displaySettings bringing the display_name over and UI just using it …

SimoneMaiwald (approved):

src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -0,0 +1,259 @@
...
+        logger.info("Getting list of participant information");
+
+        DDPInstanceDto ddpInstanceDto = participantWrapperPayload.getDdpInstanceDto()
+                .orElseThrow();
+
+        if (StringUtils.isBlank(ddpInstanceDto.getEsParticipantIndex())) {
+            throw new RuntimeException("No participant index setup in ddp_instance table for " + ddpInstanceDto.getInstanceName());
+        }
+
+        DDPInstance ddpInstance = DDPInstance.getDDPInstance(ddpInstanceDto.getInstanceName());
```
SimoneMaiwald: we need to change that to 
`DDPInstance ddpInstance = DDPInstance.getDDPInstanceWithRole(ddpInstanceDto.getInstanceName(), DBConstants.HAS_MEDICAL_RECORD_ENDPOINTS);`
 to have the role in L135 and L138 and to fix https://broadinstitute.atlassian.net/browse/DDP-6809

PR #279, opened 2021-08-17
DDP-6574 and DDP-6577

SimoneMaiwald:
[DDP-6577](https://broadinstitute.atlassian.net/browse/DDP-6577) adding the circadia Kit Management tab and circadia as an instance

[DDP-6574](https://broadinstitute.atlassian.net/browse/DDP-6574) adding triggering for kit received and kit sent dates (only trigger if ddp was not triggered for the pt already)


src/main/java/org/broadinstitute/dsm/db/dao/queue/EventDao.java
```diff
@@ -0,0 +1,58 @@
...
+    public int delete(int id) {
+        return 0;
+    }
+
+    @Override
+    public Optional<EventDto> get(long id) {
+        return Optional.empty();
+    }
+
+    public static Boolean hasTriggeredEventByEventTypeAndDdpParticipantId(Connection conn, @NotNull String eventType, @NonNull String ddpParticipantId) {
```
gmakharat: it is better to make Dao methods as instance methods and not class methods, also remove passing connection as an argument and use `TransactionWrapper.inTransaction` here to have connection

src/main/java/org/broadinstitute/dsm/db/dao/settings/EventTypeDao.java
```diff
@@ -0,0 +1,120 @@
...
+    public int delete(int id) {
+        return 0;
+    }
+
+    @Override
+    public Optional<EventTypeDto> get(long id) {
+        return Optional.empty();
+    }
+
+    public static List<EventTypeDto> getEventTypeByInstanceName(@NonNull String instanceName) {
```
gmakharat: the same about instance methods

src/main/java/org/broadinstitute/dsm/db/dao/settings/EventTypeDao.java
```diff
@@ -0,0 +1,120 @@
...
+            }
+            return execResult;
+        });
+        if (results.resultException != null) {
+            throw new RuntimeException("Error getting event types ", results.resultException);
+        }
+        return eventTypeList;
+    }
+
+    public static Optional<EventTypeDto> getEventTypeByEventTypeAndInstanceId(Connection conn, @NotNull String eventType, @NonNull String instanceId) {
```
gmakharat: the same about instance methods and passed connection.

src/main/java/org/broadinstitute/dsm/route/EventTypeRoute.java
```diff
@@ -22,7 +22,7 @@ public Object processRequest(Request request, Response response, String userId)
         String userIdRequest = UserUtil.getUserId(request);
         if (StringUtils.isNotBlank(realm)) {
             if (UserUtil.checkUserAccess(realm, userId, "participant_event", userIdRequest)) {
-                return EventType.getEventTypes(realm);
+                return EventTypeDao.getEventTypeByInstanceName(realm);
```
gmakharat: having static methods make easier to use them directly but in general, it breaks object oriented programming idea.

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -537,4 +547,19 @@ private NameValue setAdditionalValue(String additionalValue, @NonNull Patch patc
             throw new RuntimeException("DBElement not found in ColumnNameMap: " + additionalValue);
         }
     }
+
+    private void triggerParticipantEvent(DDPInstance ddpInstance, Patch patch, Value action){
+        inTransaction((conn) -> {
+            Optional<EventTypeDto> eventType = EventTypeDao.getEventTypeByEventTypeAndInstanceId(conn, action.getName(), ddpInstance.getDdpInstanceId());
```
gmakharat: having `inTransaction` in `getEventTypeByEventTypeAndInstanceId` instead of passing connection will make business logic classes to be more loosely coupled.

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -537,4 +547,19 @@ private NameValue setAdditionalValue(String additionalValue, @NonNull Patch patc
...
+            if (eventType != null && !eventType.isEmpty()) {
+                if (!EventDao.hasTriggeredEventByEventTypeAndDdpParticipantId(conn, action.getName(), patch.getParentId())) {
+                    EventUtil.triggerDDP(conn, eventType, patch.getParentId());
```
gmakharat, line 559: if this line was the key point of having connection passed to `getEventTypeByEventTypeAndInstanceId` just to avoid several connection, if think it still worth to eliminate connection from `getEventTypeByEventTypeAndInstanceId`, `EventUtil` needs to be refactored later, since, I think it does more than being utility class only.

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -549,17 +551,18 @@ private NameValue setAdditionalValue(String additionalValue, @NonNull Patch patc
     }
 
     private void triggerParticipantEvent(DDPInstance ddpInstance, Patch patch, Value action){
-        inTransaction((conn) -> {
-            Optional<EventTypeDto> eventType = EventTypeDao.getEventTypeByEventTypeAndInstanceId(conn, action.getName(), ddpInstance.getDdpInstanceId());
+            Optional<EventTypeDto> eventType = eventTypeDao.getEventTypeByEventTypeAndInstanceId(action.getName(), ddpInstance.getDdpInstanceId());
             if (eventType != null && !eventType.isEmpty()) {
```
gmakharat: I think it does not make sense to check whether `eventType` is null or not, since method `getEventTypeByEventTypeAndInstanceId` will return either empty `Optional` or filled with some data, thus you could use directly `if (eventType.isPresent())` or `eventType.ifPresent(eventTypeDto -> {//logic here})`.

gmakharat (approved):

PR #281, opened 2021-08-19
DDP 6586 | Add address tab for PanCan

ScalaChamp:
associated ticket:
https://broadinstitute.atlassian.net/browse/DDP-6586


src/main/java/org/broadinstitute/dsm/db/InstanceSettings.java
```diff
@@ -257,6 +262,7 @@ else if (!nameObject1.equals(condition.getValue())) {
                 }
             });
         });
+        System.out.println(settingsMap);
```
SimoneMaiwald: please remove

src/main/java/org/broadinstitute/dsm/db/dao/settings/InstanceSettingsDao.java
```diff
@@ -41,6 +35,17 @@
             "FROM instance_settings " +
             "WHERE ddp_instance_id = (SELECT ddp_instance_id FROM ddp_instance WHERE study_guid = ?)";
 
+    private static final String SQL_GET_HIDE_SAMPLES_TAB_BY_STUDY_GUID = "SELECT " +
+            "hide_samples_tab " +
+            "FROM instance_settings " +
+            "WHERE ddp_instance_id = (SELECT ddp_instance_id FROM ddp_instance WHERE study_guid = ?)";
```
SimoneMaiwald: maybe remove that so that it is not getting used by accident?

src/main/resources/liquibase/DDP-6586_update_has_address_tab_to_true_for_pancan.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SimoneMaiwald: can we move that file in a new folder called `CMI` thank you!

src/main/java/org/broadinstitute/dsm/db/dao/settings/InstanceSettingsDao.java
ScalaChamp: Actually I haven't added this SQL statement, it's been there, I just moved it down. 

As far as I see this statement is used in getHideSamplesTabByStudyGuid(String studyGuid) method which is does its' job properly, should we still remove this?

Or we can change this statement by changing the filtering logic of method.

e.g we can use `instance_name` instead of `study_guid` in filtering, let's wait for your answer and then we'll decide, thanks!

SimoneMaiwald: if it is used in getHideSamplesTabByStudyGuid then let's leave it. sorry I missed the usage!

src/main/java/org/broadinstitute/dsm/db/dao/settings/InstanceSettingsDao.java
```diff
@@ -92,6 +104,50 @@ public int delete(int id) {
         }
         return Optional.ofNullable((Boolean) results.resultValue);
     }
+  
+    public Optional<Boolean> getHasAddressTabByStudyInstanceName(String instanceName) {
```
gmakharat: I think making this method to return just `boolean` type value would be better, wrapping Boolean into Optional looks like just extra work, which is not needed I think,  `getBoolean` method of `ResultSet` in any case returns either true or false in any case, even if result is `SQL NULL`, as the result, returning `result.resultValue` directly from the method should be fine.
Why I think that wrapping Boolean into Optional does not worth is that, we just have either `true` or `false`, manipulating them using Optional does not give any benefit, it could even make things a little bit complicated on later use of method which returns `Optional<Boolean>`.
If I miss something, please correct.

src/main/java/org/broadinstitute/dsm/db/dao/settings/InstanceSettingsDao.java
```diff
@@ -92,6 +104,50 @@ public int delete(int id) {
...
+            return executionResult;
+        });
+        if (result.resultException != null) {
+            throw new RuntimeException("Error occured while getting has_address_tab for instance_name: "
+                    + instanceName, result.resultException);
+        }
+        return Optional.ofNullable((Boolean) result.resultValue);
+    }
+
+    public Optional<Boolean> getHasComputedObjectByStudyInstanceName(String instanceName) {
```
gmakharat: the same as above.

src/main/java/org/broadinstitute/dsm/model/elasticsearch/ESAddress.java
```diff
@@ -61,4 +64,6 @@ public String getPhone() {
     public String getRecipient() {
         return recipient;
     }
+
+    public boolean getValid() { return valid; }
```
gmakharat: usually I prefer to use `is/has` prefixes for boolean getters, I think it makes more understandable.

src/main/java/org/broadinstitute/dsm/route/DisplaySettingsRoute.java
```diff
@@ -1,5 +1,7 @@
 package org.broadinstitute.dsm.route;
 
+import com.google.gson.Gson;
+import com.sun.xml.bind.v2.runtime.output.SAXOutput;
```
gmakharat: I think this was just automatically imported by IDE and is not used in this class at all.

src/main/java/org/broadinstitute/dsm/route/DisplaySettingsRoute.java
```diff
@@ -65,14 +67,16 @@ public Object processRequest(Request request, Response response, String userId)
...
+                InstanceSettings instanceSettings = new InstanceSettings();
                 displaySettings.put("assignees", Assignee.getAssignees(realm));
                 displaySettings.put("fieldSettings", FieldSettings.getFieldSettings(realm));
                 displaySettings.put("drugs", Drug.getDrugList());
                 displaySettings.put("cancers", Cancer.getCancers());
                 displaySettings.put("activityDefinitions", ElasticSearchUtil.getActivityDefinitions(instance));
                 displaySettings.put("filters", ViewFilter.getAllFilters(userIdRequest, patchUtil.getColumnNameMap(), parent, ddpGroupId, instance.getDdpInstanceId()));
                 displaySettings.put("abstractionFields", AbstractionUtil.getFormControls(realm));
-                InstanceSettings instanceSettings = new InstanceSettings();
+                displaySettings.put("hasAddressTab", instanceSettings.getHasAddressTabByStudyInstanceName(realm));
```
gmakharat: I don't think so that it is necessary to manually put `hasAddressTab` and `hasComputedObject` to the map, `InstanceSettings` has method already called `getInstanceSettings`(Line 80) which returns all the settings by instance name(realm), then method  `getInstanceSettingsAsMap` of `InstanceSettings` will automatically extract all necessary settings from `instanceSettingsDto` with help of reflection. You would have only to change method `getInstanceSettingsByInstanceName` of `InstanceSettingsDao` by extracting two new table column from `ResultSet` .
InstanceSettingsDao, L133

src/main/java/org/broadinstitute/dsm/db/InstanceSettings.java
gmakharat: @Anzori-Ghurtchumelia

src/main/java/org/broadinstitute/dsm/db/dto/settings/InstanceSettingsDto.java
```diff
@@ -63,6 +65,13 @@
         return Optional.of(gbfShippedTriggerDSSDelivered);
     }
 
+    public Optional<Boolean> isHasAddressTab() {
```
gmakharat: either `is` or `has` will be enough, depends the context, here I would use `hasAddressTab`

src/main/java/org/broadinstitute/dsm/db/dto/settings/InstanceSettingsDto.java
```diff
@@ -63,6 +65,13 @@
...
+        return Optional.of(hasAddressTab);
+    }
+
+    public Optional<Boolean> isHasComputedObject() {
```
gmakharat: here both could fit `hasComputedObject` or `isComputedObject`

src/main/java/org/broadinstitute/dsm/db/InstanceSettings.java
```diff
@@ -93,6 +93,14 @@ public InstanceSettingsDto getInstanceSettings(String realm) {
                 .orElse(new InstanceSettingsDto.Builder().build());
     }
 
+    public boolean getHasAddressTabByStudyInstanceName(String instanceName) {
```
gmakharat: are those two method still necessary to have?

gmakharat (approved):

SimoneMaiwald (approved):

PR #288, opened 2021-09-15
DDP-6535 | fix search mechanism for AGREEMENT type columns

ScalaChamp:


gmakharat (approved):

src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -57,6 +57,8 @@
     public static String CHECKBOX = "CHECKBOX";
     public static String COMPOSITE = "COMPOSITE";//ES type
     public static String JSON_ARRAY = "JSONARRAY";//Sample result
+    public static String AGREEMENT = "AGREEMENT";//Sample result
```
SimoneMaiwald: "//Sample result" is not correct here

SimoneMaiwald (approved):

src/main/java/org/broadinstitute/dsm/model/Filter.java
ScalaChamp: @SimoneMaiwald correcting it right now

SimoneMaiwald (approved):

PR #289, opened 2021-09-16
DDP-4672 | make First/Last names and shortID searchable through exact and non-exa…

ScalaChamp:
…ct matches

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
@@ -133,7 +133,7 @@ public class ElasticSearchUtil {
         }
     }
 
-    public static void fetchFieldMappings() {
+    private static void fetchFieldMappings() {
         GetMappingsRequest request = new GetMappingsRequest();
         request.indices(PARTICIPANTS_STRUCTURED_ANY);
         try {
--- a/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
@@ -98,6 +98,7 @@ public class ElasticSearchUtil {
     public static final String PROFILE_LEGACYALTPID = "profile.legacyAltPid";
     public static final String WORKFLOWS = "workflows";
     public static final String EMAIL_FIELD = "email";
+    public static final String NORMALIZE = ".normalize";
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -111,11 +128,24 @@ public static synchronized void initClient() {
                         TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.ES_USERNAME),
                         TransactionWrapper.getSqlFromConfig(ApplicationConfigConstants.ES_PASSWORD));
             } catch (MalformedURLException e) {
-                throw new RuntimeException("Error while initialize ES client", e);
+                throw new RuntimeException("Error while initializing ES client", e);
             }
         }
     }
 
+    public static void fetchFieldMappings() {
```
gmakharat: I think having this method as `private` would be better, since it is used only in this class and I believe outside of this class does not need to know about it.

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -1405,26 +1348,52 @@ private static void valueQueryBuilder(@NonNull BoolQueryBuilder finalQuery, @Non
...
+
+
+    private static String getFieldTypeByFieldName(String name) {
+        String anyStudy = getAnyStudy();
+        String fields = getFieldsAsString(anyStudy);
+        String[] fieldsArray = name.split(DOT_SEPARATOR);
+        String outerField = fieldsArray[OUTER_FIELD_INDEX];
+        String innerField = fieldsArray[INNER_FIELD_INDEX];
+        Gson gson = new Gson();
+        HashMap fieldsMap = gson.fromJson(fields, HashMap.class);
```
gmakharat, line 1385: maybe having `Map` instead of `HashMap` on variable declaration would be good with types.

ScalaChamp: Unfortunately the casting mechanism does not work with LinkedTreeMap of Gson, I'm afraid I'll leave this as it is.

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
ScalaChamp: correcting it

gmakharat (approved):

SimoneMaiwald (approved):

PR #290, opened 2021-09-16
DDP-6651 epoch date filters

zyme:
Although DDP-6651 mentions a lot of different date issues, this PR focuses on two patterns: dates that are stored as epoch and dates that are stored as strings.

The approach here is to introduce a new annotation (`@DbDateConversion`) that is picked up by the filter to use the appropriate transformation into SQL.  There is a special case for "day equals" vs. raw date format match, and if a column isn't annotated, no conversion is done.

Start the PR looking at the date [conversion annotation](https://github.com/broadinstitute/ddp-study-manager/pull/290/files#r710414085) and its usages along side of `@ColumnName`, where the [converter does the work](https://github.com/broadinstitute/ddp-study-manager/pull/290/files#r710413627).  Then dive into the method that uses the [filtering to modify db queries in the Filter class](https://github.com/broadinstitute/ddp-study-manager/pull/290/files#r710412118).

There were some related improvements that I was able to make along the way, and those are scattered throughout the code changes.

This should make date queries work for many columns, include kit dates, medical record dates, onc history, and kit requests.  Because the storage of dynamic forms is quite different, I was not able to include a solution for those here.

I also noticed that although the UI seems to return dates in a "local" timezone, it's not clear to me how this timezone is selected.  The backend meanwhile is assuming UTC in the database, so there is room for confusion here when returned dates appear to fall outside the desired range.


src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -268,7 +266,35 @@ else if (filter.getFilter1() != null && filter.getFilter2().getValue() != null &
         return finalQuery;
     }
 
+    /**
+     * Uses the appropriate date converter (if given) to write SQL that can
+     * compare either exact dates or "in the day" dates.
+     */
+    private static String generateDateComparisonSql(Filter filter, DBElement dbElement, String comparison, Object arg) {
```
zyme, line 274: Breadcrumb for SQL generation link...

src/main/java/org/broadinstitute/dsm/db/structure/SqlDateConverter.java
```diff
@@ -0,0 +1,30 @@
+package org.broadinstitute.dsm.db.structure;
+
+import java.time.Instant;
+
+public enum SqlDateConverter {
```
zyme, line 5: Link from PR...

src/main/java/org/broadinstitute/dsm/db/structure/DbDateConversion.java
```diff
@@ -0,0 +1,16 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+/**
+ * Defines how to convert the column into a date during date filtering
+ */
+public @interface DbDateConversion  {
```
zyme, line 14: Link from PR

src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -941,7 +945,7 @@ else if (word.equals(Filter.MINUS_SIGN)) {
                 filter.setSelectedOptions(selectedOptions.toArray(b));
                 filter.setType(type == null ? Filter.TEXT : type);
                 if (isValidDate(value, false)) {
-                    type = Filter.DATE;
```
zyme, line 944: Seemed like a latent bug--this format is a date.

src/main/java/org/broadinstitute/dsm/util/PatchUtil.java
```diff
@@ -104,7 +106,14 @@ private void getColumnNames(Class<?> obj) {
                     if (StringUtils.isNotBlank(columnPrefix)) {
                         fieldKey = columnPrefix.concat("_").concat(field.getName());
                     }
-                    columnNameMap.put(fieldKey, new DBElement(tableName, tableAlias, primaryKey, column.value()));
+                    DbDateConversion[] dbDateConversions = field.getAnnotationsByType(DbDateConversion.class);
+                    SqlDateConverter dateConverter = null;
+                    if (dbDateConversions != null && dbDateConversions.length == 1) {
```
gmakharat: `getAnnotationsByType` anyway returns an empty array if there is no such annotation, so I think checking it on null is pointless, also since `DbDateConversion` is not `Repeatable`, using `getAnnotation` instead of `getAnnotationsByType` could be more convenient.

src/main/java/org/broadinstitute/dsm/db/structure/DBElement.java
```diff
@@ -9,11 +14,89 @@
...
+        }
+    }
+
+    /**
+     * Use this for simple yyyy-mm-dd string columns
+     */
+    public static class StringDayConverter implements DateConverter {
+
+        public String convertArgToSql(Instant arg) {
+            return "STR_TO_DATE(" + "'" + DateTimeFormatter.ofPattern("yyyy-MM-dd").withZone(ZoneId.of("UTC")).format(arg) + "'" + ",'%Y-%m-%d')";
```
gmakharat: maybe using already existing `DateTimeFormatter.ISO_LOCAL_DATE` format could be more cleaner?

src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -268,6 +266,33 @@ else if (filter.getFilter1() != null && filter.getFilter2().getValue() != null &
...
+        String column = filter.getColumnName(dbElement);
+        SqlDateConverter dateConverter = dbElement.getDateConverter();
+
+        Instant instant = null;
+        try {
+            instant = new SimpleDateFormat("yyyy-MM-dd").parse(arg.toString()).toInstant().atZone(ZoneId.of("UTC")).toInstant();
```
gmakharat: `SimpleDateFormat` is quite old already, I think it is better idea to use date related things from `time` package, for example: 
`LocalDate.parse(arg.toString(), DateTimeFormatter.ISO_LOCAL_DATE).atStartOfDay().toInstant(ZoneOffset.UTC)` should be better.

src/main/java/org/broadinstitute/dsm/util/PatchUtil.java
```diff
@@ -104,7 +106,14 @@ private void getColumnNames(Class<?> obj) {
                     if (StringUtils.isNotBlank(columnPrefix)) {
                         fieldKey = columnPrefix.concat("_").concat(field.getName());
                     }
-                    columnNameMap.put(fieldKey, new DBElement(tableName, tableAlias, primaryKey, column.value()));
+                    DbDateConversion dbDateConversion = field.getAnnotation(DbDateConversion.class);
+                    SqlDateConverter dateConverter = null;
+                    if (dbDateConversion != null) {
```
gmakharat: in any case, does not matter `field.getAnnotation(DbDateConversion.class)` returned actual value or null, we pass dateConverter to `DBElement` constructor, what if we can eliminate if check for null, something like that: 
`SqlDateConverter dateConverter = field.getAnnotation(DbDateConversion.class).value();`

zyme: Won't this throw an NPE if there's no annotation for the field?

zyme: We could pass `DbDateConversion` to the `DBElement` constructor and do the null check there--that would make calling code a bit cleaner.

gmakharat: oh good catch, my bad sorry, about null checks usually I prefer to use `Optional` getters for the fields, something like:
```
     public class DBElement {
       private SqlDateConverted dateConverter;
       
       public Optional<SqlDateConverted> getDateConverter() {
            return Optional.ofNullable(this.dateConverter);
       }
     }
```
and caller will have to handle it with methods of `Optional` which provides much cleaner code from my point of view.

gmakharat (approved):

PR #291, opened 2021-09-20
DDP-6651 Date filtering on participant list fields

zyme:
The backend was not recognizing `ADDITIONALVALUE` fields, resulting in them being passed through to elastic instead of queried directly in the database.  The fix here was to pass through the `parent` name from the frontend through to the filter parsing, and then use this to recognize that a field should be queried from within the database.

Along the way, fixed an NPE that Gene reported.


src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -964,6 +965,10 @@ else if (!Filter.CHECKBOX.equals(filter.type)) {
                     if (path != null && !f2) {//additional field
                         filter.setFilter2(new NameValue(path, ""));
                     }
+                    if (f1 && !f2 && Filter.DATE.equals(filter.type) && filter.isRange()) {
```
zyme, line 968: When there's a range set and there's no end date--as was seen in at least one saved filter--let's just look waaaay in the future instead of requiring the user to set "and less than today"

gmakharat (approved):

PR #292, opened 2021-09-20
DDP-6936: proxy information

SimoneMaiwald:
filter by proxy

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
+++ b/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
@@ -262,8 +262,7 @@ public class ViewFilter {
         }
         else {
             String queryCondition = "";
-            if (filter != null && filter.getParentName() != null &&
-                    filter.getParticipantColumn().getTableAlias() != null && ElasticSearchUtil.PROXY.equals(filter.getParticipantColumn().getTableAlias())) {
+            if (isFilterForProxyInformation(filter)) {
                 if (queryConditions.containsKey(ElasticSearchUtil.PROXY)) {
                     queryCondition = queryConditions.get(ElasticSearchUtil.PROXY);
                 }
@@ -277,6 +276,14 @@ public class ViewFilter {
                 queryConditions.put(ElasticSearchUtil.ES, queryCondition.concat(Filter.getQueryStringForFiltering(filter, null)));
             }
         }
+    }
+
+    private static boolean isFilterForProxyInformation(Filter filter) {
+        if (filter != null && filter.getParentName() != null &&
+                filter.getParticipantColumn().getTableAlias() != null && ElasticSearchUtil.PROXY.equals(filter.getParticipantColumn().getTableAlias())) {
+            return true;
+        }
+        return false;
     }
 
     public static List<ViewFilter> getAllFilters(@NonNull String userId,
--- a/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
+++ b/src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
@@ -15,6 +15,7 @@ import org.broadinstitute.dsm.model.NameValue;
 import org.broadinstitute.dsm.model.ParticipantColumn;
 import org.broadinstitute.dsm.model.TissueList;
 import org.broadinstitute.dsm.statics.DBConstants;
+import org.broadinstitute.dsm.util.ElasticSearchUtil;
 import org.broadinstitute.dsm.util.PatchUtil;
 import org.broadinstitute.dsm.util.SystemUtil;
 import org.slf4j.Logger;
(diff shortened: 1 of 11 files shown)


src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -261,10 +261,20 @@ public static void addQueryCondition(@NonNull Map<String, String> queryCondition
         }
         else {
             String queryCondition = "";
-            if (queryConditions.containsKey("ES")) {
-                queryCondition = queryConditions.get("ES");
+            if (filter != null && filter.getParentName() != null &&
```
gmakharat: I think extracting this long if condition into method with meaningful name would be useful.

src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -261,10 +261,20 @@ public static void addQueryCondition(@NonNull Map<String, String> queryCondition
...
+                    filter.getParticipantColumn().getTableAlias() != null && "proxy".equals(filter.getParticipantColumn().getTableAlias())) {
+                if (queryConditions.containsKey("proxy")) {
```
gmakharat: having `proxy` string as a constant in `Filter` would be convenient, for later usage in somewhere else as well.

src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -261,10 +261,20 @@ public static void addQueryCondition(@NonNull Map<String, String> queryCondition
...
+                    queryCondition = queryConditions.get("proxy");
+                }
+                queryConditions.put("proxy", queryCondition.concat(Filter.getQueryStringForFiltering(filter, null)));
+
+            }
+            else {
+                if (queryConditions.containsKey("ES")) {
```
gmakharat: there is a constant for this `ElasticSearchUtils.ES`.

src/main/java/org/broadinstitute/dsm/model/elasticsearch/ElasticSearchParticipantDto.java
```diff
@@ -22,6 +22,7 @@
     private String status;
     private ESDsm dsm;
     private String ddp;
+    private List<String> governedUsers;
```
gmakharat, line 27: `governedUsers` is the exact name how it comes from ElasticSearch? asking because converting source map of ElasticSearch to the object of `ElasticSearchParticipantDto` is done by using `GSON` and field names needs to match names which come from ElasticSearch's source map

src/main/java/org/broadinstitute/dsm/model/elasticsearch/ElasticSearchParticipantDto.java
```diff
@@ -67,6 +68,10 @@
         return Optional.ofNullable(dsm);
     }
 
+    public Optional<List<String>> getGovernedUsers() {
```
gmakharat: I think it is better idea to return empty list from getters, if field is type of `List`, like:
```
public List<String> getGovernedUsers() {
    if (this.governedUsers == null) {
        this.governedUsers = Collections.emptyList();
    }
    return this.governedUsers;
}
```
I see that there are other getters which return `List` types wrapped in `Optional`, but I think it will be good idea if we change them as well, since returning empty list instead of wrapping it in `Optional` will be more convenient for use and also it will help us to avoid NPEs as well.

src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -116,6 +116,10 @@ else if (DBConstants.DDP_ABSTRACTION_ALIAS.equals(source)) {
                     abstractionActivities = AbstractionActivity.getAllAbstractionActivityByRealm(ddpInstance.getName(), filters.get(source));
                     participantIdsToFetch = new ArrayList<>(abstractionActivities.keySet());
                 }
+                else if ("proxy".equals(source)) {
```
gmakharat: "proxy" here will be changed as well once it will be extracted as constant.

src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -165,6 +169,16 @@ else if (DBConstants.DDP_ABSTRACTION_ALIAS.equals(source)) {
         }
     }
 
+    List<String> getGovernedUserIdsFromElasticList(List<ElasticSearchParticipantDto> proxyData) {
+        List<String> participantGovernedUserIds = new ArrayList<>();
+
+        proxyData.stream()
+                .filter(esGovernedUserData -> esGovernedUserData.getGovernedUsers().orElse(Collections.emptyList()).size() > 0)
```
gmakharat: if we have `getGovernedUsers()` to return empty list instead of wrapping it in `Optional` it should be much more convenient, since we won't have to do extra work like `orElse`, `Optional` is powerful if field type is not `Collection`.

src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
SimoneMaiwald: Good point! I will add it to ElasticSearchUtil because it is a ES object

src/main/java/org/broadinstitute/dsm/model/elasticsearch/ElasticSearchParticipantDto.java
SimoneMaiwald: yes, that is the ES object name

gmakharat (dismissed):

src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -165,6 +170,25 @@ else if (DBConstants.DDP_ABSTRACTION_ALIAS.equals(source)) {
...
+
+        proxyData.stream()
+                .filter(esGovernedUserData -> !esGovernedUserData.getGovernedUsers().isEmpty())
+                .forEach(esGovernedUserData ->
+                        participantGovernedUserIds.addAll(esGovernedUserData.getGovernedUsers()));
+        return participantGovernedUserIds;
+    }
+
+    private List<String> combineListWithMatchingParticipantIds(List<String> participantIdsToFetch, Set<String> list) {
+        if (participantIdsToFetch.isEmpty()) {
```
gmakharat: I would change this `if/else` to ternary operator:
```
   return participantIdsToFetch.isEmpty() 
             ? //...
             : //...;
```

src/main/java/org/broadinstitute/dsm/model/participant/ParticipantWrapper.java
```diff
@@ -165,6 +170,25 @@ else if (DBConstants.DDP_ABSTRACTION_ALIAS.equals(source)) {
...
+            return new ArrayList<>(list);
+        }
+        else {
+            return participantIdsToFetch.stream().filter(new HashSet<>(list)::contains).collect(Collectors.toList());
```
gmakharat: is it necessary to wrap list into `HashSet` to use `contains` method? Maybe `filter(list::contains)` could be enough as well?

PR #293, opened 2021-09-20
DDP-6681 - fix filtering of numeric values (like ages)

evgeniipr:
Fix filtering of numeric values (like ages): fix a case when it is specified only one limit of range or both limits.
For example fixed generation of ES queries from such expressions:
`AND PREQUAL.SELF_CURRENT_AGE >= 20 AND PREQUAL.SELF_CURRENT_AGE <= 46 AND PREQUAL.SELF_CURRENT_AGE IS NOT NULL`
`AND PREQUAL.SELF_CURRENT_AGE >= 20 AND PREQUAL.SELF_CURRENT_AGE IS NOT NULL`
`AND PREQUAL.SELF_CURRENT_AGE <= 46 AND PREQUAL.SELF_CURRENT_AGE IS NOT NULL`

Without this fix the generated ES queries for numerics are invalid. Sometimes queries with numeric values caused errors and NPEs.

The fix is done as a "workaround" (trying not to touch the existing code) but in reality it would be good to do a global refactoring of filter expressions parsing and ES queries generation.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
@@ -827,6 +827,9 @@ public class ElasticSearchUtil {
         String[] filters = filter.split(Filter.AND);
         BoolQueryBuilder finalQuery = new BoolQueryBuilder();
 
+        Map<String, String> queryPartsMap = new HashMap<>();  // store query path pointing associated with field name (to be compared)
+        NestedQueryBuilder parentNestedOfRangeBuilder = null; // store NestedQueryBuilder which is a parent of RangeBuilder
+
         for (String f : filters) {
             if (StringUtils.isNotBlank(f) && f.contains(DBConstants.ALIAS_DELIMITER)) {
                 if (f.contains(Filter.EQUALS) || f.contains(Filter.LIKE)) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -1012,6 +1033,23 @@ else if (nameValue[0].startsWith(INVITATIONS)) {
                                 catch (ParseException e) {
                                     logger.error("range was not date. user entered: " + userEntered);
                                 }
+                            } else if (StringUtils.isNumeric(userEntered)) { // separately process expressions with numbers
+                                QueryBuilder tmpBuilder = findQueryBuilderForFieldName(finalQuery, surveyParam[1], queryPartsMap);
```
gmakharat: I think it will be good idea to extract content of this if block as a method, since nearly the very same code is written above, from L933 - L949, to avoid code duplication.

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -1062,6 +1100,23 @@ else if (nameValue[0].startsWith(INVITATIONS)) {
...
+
+                                // search for queries for the same field:
+                                // if found and it is Range and it was inserted into Nested block then remove
+                                // this Range from former place and add it into must() block together with this
+                                // `activities.questionsAnswers.stableId`==[field_name] query.
+                                QueryBuilder tmpBuilder = findQueryBuilderForFieldName(finalQuery, surveyParam[1], queryPartsMap);
+                                if (tmpBuilder instanceof RangeQueryBuilder && parentNestedOfRangeBuilder != null) {
+                                    activityAnswer.must(tmpBuilder);
+                                    BoolQueryBuilder finalQueryCopy = new BoolQueryBuilder();
+                                    for (Iterator<QueryBuilder> iterator = finalQuery.must().iterator(); iterator.hasNext(); ) {
```
gmakharat: I think using for each loop instead of old version of usage of for with `Iterator` is more flexible and cleaner.

evgeniipr: I agree: I tried to use forEach loop but encountered a problem that it gives error regarding the variable `parentNestedOfRangeBuilder` complaining that it is not final or effectively final. In case of iterator it doesn't complain.

gmakharat: Strange, you are not changing value of `parentNestedOfRangeBuilder`, even more we don't have lambda expression or anonymous class to have capturing problem, did you try like this construction: 
```
for (QueryBuilder builder : finalQuery.must()) {
    if (!builder.equals(parentNestedOfRangeBuilder)) {
        finalQueryCopy.must(builder);
    }
}
```

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
evgeniipr: Extracted the code.

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
evgeniipr: Extracted the code to a method and then was able to use each loop instead of iterator.

gmakharat (approved):

PR #296, opened 2021-09-22
DDP-6651 Check a date range inclusively

zyme:
One last (ha!) bit for DDP-6651: when we've parsed two dates for a range comparison, include all times in both days, as well as all times in all days inbetween.  Previously, the start time of the start of the range was used properly, but the end of the range stopped at the first millis of the day instead of the last millis of the day.

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/model/Filter.java
+++ b/src/main/java/org/broadinstitute/dsm/model/Filter.java
@@ -298,14 +298,10 @@ public class Filter {
             Instant instant = null;
             try {
                 LocalDate date = LocalDate.parse(arg.toString(), DateTimeFormatter.ISO_LOCAL_DATE);
-                if (useEndOfday) {
-                    instant = date.atTime(LocalTime.MAX).toInstant(ZoneOffset.UTC);
-                } else {
-                    instant = date.atStartOfDay().toInstant(ZoneOffset.UTC);
-                }
+                instant = useEndOfday ? date.atTime(LocalTime.MAX).toInstant(ZoneOffset.UTC) : date.atStartOfDay().toInstant(ZoneOffset.UTC);
             } catch (DateTimeParseException e) {
                 // might be an epoch time in an older saved filter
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -281,15 +282,27 @@ private static boolean isNotEmpty(NameValue filter) {
...
+    private static String generateDateComparisonSql(Filter filter, DBElement dbElement, String comparison, Object arg, boolean useEndOfday) {
         String column = filter.getColumnName(dbElement);
         SqlDateConverter dateConverter = null;
         if (dbElement != null) {
             dateConverter = dbElement.getDateConverter();
             Instant instant = null;
             try {
-                instant = LocalDate.parse(arg.toString(), DateTimeFormatter.ISO_LOCAL_DATE).atStartOfDay().toInstant(ZoneOffset.UTC);
+                LocalDate date = LocalDate.parse(arg.toString(), DateTimeFormatter.ISO_LOCAL_DATE);
+                if (useEndOfday) {
```
gmakharat: I would use ternary operator here, but I think it is personal preference.

src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -281,15 +282,27 @@ private static boolean isNotEmpty(NameValue filter) {
...
+                    instant = date.atTime(LocalTime.MAX).toInstant(ZoneOffset.UTC);
+                } else {
+                    instant = date.atStartOfDay().toInstant(ZoneOffset.UTC);
+                }
             } catch (DateTimeParseException e) {
                 // might be an epoch time in an older saved filter
                 instant = Instant.ofEpochMilli(Long.valueOf(arg.toString()));
```
gmakharat: I think using `Long.parseLong()` method would be useful to avoid extra operation by using `valueOf`.

gmakharat (approved):

gmakharat (approved):

SimoneMaiwald (approved):

PR #301, opened 2021-09-27
DDP-4762 | fix search mechanism for "status"

ScalaChamp:


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -1516,14 +1516,45 @@ private static String getFieldTypeByFieldName(String name) {
...
+        finalField = gson.fromJson(String.valueOf(finalField.get(field)), HashMap.class);
+        return finalField;
+    }
+
+    private static String getFieldTypeByArrayLength(String[] fieldsArray, String outerField, Gson gson, HashMap propertiesMap) {
+        if (fieldIsNested(fieldsArray)) {
+            String innerField = fieldsArray[INNER_FIELD_INDEX];
+            HashMap outerFieldsMap = gson.fromJson(String.valueOf(propertiesMap.get(outerField)), HashMap.class);
+            HashMap outerFieldsPropertiesMap = gson.fromJson(String.valueOf(outerFieldsMap.get(PROPERTIES)), HashMap.class);
+            return (String) gson.fromJson(String.valueOf(outerFieldsPropertiesMap.get(innerField)), HashMap.class).get(TYPE);
```
gmakharat: is there chance that result of `outerFieldsPropertiesMap.get(innerField)` could be null? if so, calling `.get(TYPE)` will throw NPE

ScalaChamp: @gmakharat it will only happen if ES index will be empty, like for example the `participants_structured.json` will be empty.

BTW please take a look at the second commit, this one also works, but in the second commit I found the more `programmatic` way of doing the same, which is better because the field depth can be dynamic, like this:

data
data.field
data.field.innerfield
data.field.innerfield.innerfield

etc

and the second commit solves this

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -1518,8 +1518,28 @@ private static String getFieldTypeByFieldName(String name) {
         String outerField = fieldsArray[OUTER_FIELD_INDEX];
         Gson gson = new Gson();
         HashMap fieldsMap = gson.fromJson(fields, HashMap.class);
-        HashMap propertiesMap = gson.fromJson(String.valueOf(fieldsMap.get(PROPERTIES)), HashMap.class);
-        return getFieldTypeByArrayLength(fieldsArray, outerField, gson, propertiesMap);
+        HashMap propertiesMap = getField(gson, fieldsMap, PROPERTIES);
+        HashMap finalField = getField(gson, propertiesMap, outerField);
+        for (String field : Arrays.copyOfRange(fieldsArray, INNER_FIELD_INDEX, fieldsArray.length)) {
```
ScalaChamp, line 1529: @gmakharat what do you think if I add an if statement here that checks the length of `fieldsArray`? 
it will be clearer to understand that if the length of `fieldsArray` is greater than one it means that we are looking for a nested field, otherwise we are looking for the outer field in the root and there's no need to go in the loop.

Technically it does not change anything but it will be easier to reason about it

gmakharat: sounds good, you can use already existing method `fieldIsNested` I think, but I would rename it with `isFieldNested`.

gmakharat (approved):

PR #309, opened 2021-10-08
Refactor of PatchRoute

gmakharat:
**Things have done:**

- Introduced OO principles to separate responsibilities into different classes.
- Created `BasePatch` abstract class as a parent of different classes, which are serving common feature.
- Introduced Template method design pattern in `BasePatch` to extract common code to avoid code duplication.
- Created abstract methods in `BasePatch` to specify what children should do to patch data.
- Introduced factory method to hide low level details about creation of special instance of `BasePatch`.

P.S code was written using pair programming with @Anzori-Ghurtchumelia.


src/main/java/org/broadinstitute/dsm/model/patch/AbstractionPatch.java
```diff
@@ -0,0 +1,61 @@
...
+        return Map.of(PRIMARY_KEY_ID, primaryKeyId);
+    }
+
+    @Override
+    Optional<Object> processEachNameValue(NameValue nameValue, DBElement dbElement) {
+        if (StringUtils.isBlank(primaryKeyId)) {
+            primaryKeyId = AbstractionWrapper.createNewAbstractionFieldValue(patch.getParentId(), patch.getFieldId(), patch.getUser(), nameValue, dbElement);
+        }
+        if (!Patch.patch(primaryKeyId, patch.getUser(), nameValue, dbElement)) {
+            throw new RuntimeException("An error occurred while attempting to patch ");
```
SimoneMaiwald: Could we add more info here? not the value in case it contains PHI but user and field

src/main/java/org/broadinstitute/dsm/model/patch/MedicalRecordPatch.java
```diff
@@ -0,0 +1,100 @@
...
+import org.broadinstitute.dsm.db.OncHistoryDetail;
+import org.broadinstitute.dsm.db.structure.DBElement;
+import org.broadinstitute.dsm.model.NameValue;
+import org.broadinstitute.dsm.model.Patch;
+import org.broadinstitute.dsm.statics.DBConstants;
+import org.broadinstitute.dsm.util.MedicalRecordUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MedicalRecordPatch extends BasePatch {
```
SimoneMaiwald: MedicalRecordPatch or OncHistoryDetailPatch?

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -68,501 +35,31 @@ public PatchRoute(@NonNull NotificationUtil notificationUtil, @NonNull PatchUtil
...
-                            if (dbElement == null) {
-                                throw new RuntimeException("DBElement not found in ColumnNameMap: " + patch.getNameValue().getName());
-                            }
-                            Patch.patch(String.valueOf(participantId), patch.getUser(), patch.getNameValue(), dbElement);
-                            return new Result(200, gson.toJson(Map.of(PARTICIPANT_ID, String.valueOf(participantId))));
-                        }
-                    }
-                }
-                throw new RuntimeException("Id and parentId was null");
+                Patch patch = GSON.fromJson(requestBody, Patch.class);
```
SimoneMaiwald, line 46: I love how clean that Route is now

src/main/java/org/broadinstitute/dsm/model/patch/AbstractionPatch.java
gmakharat: we checked and false value from `Patch.patch` will never be returned, method `Patch.patch` itself throws several exception which already contains meaningful information for developer ,thus we decided to remove if check for `!Patch.patch`.

PR #310, opened 2021-10-12
DDP-7043: multiple date bugs

SimoneMaiwald:


src/main/java/org/broadinstitute/dsm/db/ViewFilter.java
```diff
@@ -994,10 +995,10 @@ else if (!Filter.CHECKBOX.equals(filter.type)) {
...
                         filter.setFilter2(new NameValue(path, ""));
                     }
-                    if (f1 && !f2 && Filter.DATE.equals(filter.type) && filter.isRange()) {
-                        // set max date to very far in the future
-                        filter.setFilter2(new NameValue(filter.getFilter1().getName(), LocalDateTime.now().plusYears(10).format(DateTimeFormatter.ISO_LOCAL_DATE)));
-                    }
+//                    if (f1 && !f2 && Filter.DATE.equals(filter.type) && filter.isRange()) {
+//                        // set max date to very far in the future
+//                        filter.setFilter2(new NameValue(filter.getFilter1().getName(), LocalDateTime.now().plusYears(10).format(DateTimeFormatter.ISO_LOCAL_DATE)));
+//                    }
```
gmakharat, line 1001: Do we need commented out lines?

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -1035,7 +1041,13 @@ public static String getPreferredLanguage(@NonNull Map<String, Map<String, Objec
                                 }
                             } else if (StringUtils.isNumeric(userEntered)) { // separately process expressions with numbers
                                 NestedQueryBuilder query = addRangeLimitForNumber(
-                                        Filter.SMALLER_EQUALS, surveyParam[1], userEntered, finalQuery, queryPartsMap);
+                                        Filter.SMALLER_EQUALS, surveyParam[1].trim(), userEntered, finalQuery, queryPartsMap, ACTIVITIES_QUESTIONS_ANSWER_ANSWER);
+                                if (query != null) {
+                                    parentNestedOfRangeBuilderOfNumbers = query;
+                                }
```
gmakharat, line 1047: The same piece of code is presented on line 959, I think it will be good idea to extract this piece of code as a method to avoid code duplication.

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -1035,7 +1041,13 @@ public static String getPreferredLanguage(@NonNull Map<String, Map<String, Objec
...
+                            } else if (userEntered.trim().matches("\\d{4}-\\d{1,2}-\\d{1,2}")) { // separately process expressions with numbers
```
gmakharat, line 1048: the expression of this `else if` statement is used on line 964, extracting it as a method with meaningful name would be useful for better understanding what we check and to avoid code duplication as well. Method name could be something like `isDate` or something else with prefix `is/has`.

src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
SimoneMaiwald: It is using Filter.LARGER_EQUALS in line 959

PR #312, opened 2021-10-22
DDP-7034: moving date fixes from prod branch into develop

SimoneMaiwald:
changes and comments from https://github.com/broadinstitute/ddp-study-manager/pull/310

### DIFF
--- a/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
+++ b/src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
@@ -133,7 +133,7 @@ public class ElasticSearchUtil {
         }
     }
 
-    private static void fetchFieldMappings() {
+    private static synchronized void fetchFieldMappings() {
         GetMappingsRequest request = new GetMappingsRequest();
         request.indices(PARTICIPANTS_STRUCTURED_ANY);
         try {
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/dsm/util/ElasticSearchUtil.java
```diff
@@ -1013,7 +1019,13 @@ public static String getPreferredLanguage(@NonNull Map<String, Map<String, Objec
                                 }
                             } else if (StringUtils.isNumeric(userEntered)) { // separately process expressions with numbers
                                 NestedQueryBuilder query = addRangeLimitForNumber(
-                                        Filter.SMALLER_EQUALS, surveyParam[1], userEntered, finalQuery, queryPartsMap);
+                                        Filter.SMALLER_EQUALS, surveyParam[1].trim(), userEntered, finalQuery, queryPartsMap, ACTIVITIES_QUESTIONS_ANSWER_ANSWER);
+                                if (query != null) {
+                                    parentNestedOfRangeBuilderOfNumbers = query;
+                                }
```
gmakharat: The code from line 1021 to 1025(inclusive) is pretty the same for below `else if`, difference is only `Filter.SMALLER_EQUALS` and `questionAnswerType`, the same code is used above as well on line 938, I would extract it as a method and would help to eliminate code duplication which is presented in 4 places right now:
1. 937-941
2. 943-947
3. 1021-1025
4. 1027-1031

Usage of IntelliJ IDEA's `extract method` refactoring tool could be very useful here to find all the duplication and change with extracted method.

SimoneMaiwald: OMG, I didn't know about that tool in IntelliJ 💯

PR #313, opened 2021-10-29
DDP-7010: changes for mercury dummy endpoint

SimoneMaiwald:


src/main/java/org/broadinstitute/dsm/route/ClinicalKitsRoute.java
```diff
@@ -64,11 +61,14 @@ private ClinicalKitDto getClinicalKit(String kitLabel) {
...
         maybeKitInfo.ifPresent(kitInfo -> {
-            logger.info("Creating clinical kit to return to GP "+kitLabel);
+            logger.info("Creating clinical kit to return to GP " + kitLabel);
             clinicalKit.setCollaboratorParticipantId(kitInfo.getCollaboratorParticipantId());
             clinicalKit.setSampleId(kitInfo.getCollaboratorSampleId());
             clinicalKit.setMaterialType(kitInfo.getMaterialInfo());
             clinicalKit.setVesselType(kitInfo.getReceptacleName());
+            clinicalKit.setSampleType(kitInfo.getKitType());
+            clinicalKit.setMfBarcode(kitLabel);
+            clinicalKit.setCollectionDate("01/31/2021"); //TODO needs to get replaced to real values after we add these values to DSM
```
SimoneMaiwald, line 71: we do not track that information yet in DSM, there is a request to add it and before the clinical kits will go live we need to add that, but for GP mercury testing hardcoding seems fine

src/main/java/org/broadinstitute/dsm/db/dto/kit/ClinicalKitDto.java
```diff
@@ -46,23 +39,40 @@
...
+    @SerializedName("kit_label")
+    String mfBarcode;
 
     public ClinicalKitDto(){}
 
     public void setSampleType(String kitType){
         switch (kitType.toLowerCase()){
             case "saliva":
-                this.sampleType = NORMAL;
+                this.sampleType = "Normal";
```
pegahtah, line 53: I didn't understand why we changed them... ( not that the change. breaks anything, just didn't understand the reason behind it ☺️ )

pegahtah (approved):

PR #317, opened 2021-11-15
GoogleAnalytics_prod

pegahtah:


src/main/java/org/broadinstitute/dsm/analytics/GoogleAnalyticsMetrics.java
```diff
@@ -0,0 +1,29 @@
...
+//    public static final String EVENT_ACTION_USER_REGISTRATION = "register-user";
+//    public static final String EVENT_LABEL_USER_REGISTRATION = "registration"; we may need it for FON
+
+    public static final String EVENT_CATEGORY_USER_LOGIN = "user-login";
+    public static final String EVENT_ACTION_USER_LOGIN = "user-logged-in";
+    public static final String EVENT_LABEL_USER_LOGIN = "login"; //studyGuid appended
+
+    public static final String  EVENT_CATEGORY_PARTICIPANT_LIST = "participant-list";
+    public static final String EVENT_ACTION_PARTICIPANT_LIST = "participant-list-load-time";
+    public static final String EVENT_LABEL_PARTICIPANT_LIST = "participant-list-load-time";//studyGuid appended
```
SimoneMaiwald: same string? can we lose one then?

src/main/java/org/broadinstitute/dsm/analytics/GoogleAnalyticsMetrics.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public static final String  EVENT_CATEGORY_SERVER_START= "server-start";
+    public static final String EVENT_ACTION_SERVER_START = "server-start";
+    public static final String EVENT_LABEL_SERVER_START = "server-start";
```
SimoneMaiwald: same strings, do we really need it 3 times?

src/main/java/org/broadinstitute/dsm/analytics/GoogleAnalyticsMetrics.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public static final String EVENT_CATEGORY_PATCH_DATA = "patch-data";
+    public static final String EVENT_ACTION_PATCH_DATA = "patch-data-answers";
+    public static final String EVENT_LABEL_PATCH_DATA = "patch-answers"; //studyGuid & original page appended
+
+    public static final String EVENT_CATEGORY_TISSUE_LIST = "tissue-list";
+    public static final String EVENT_ACTION_TISSUE_LIST = "tissue-list-loaded";
+    public static final String EVENT_LABEL_TISSUE_LIST = "tissue-list-loaded";//studyGuid appended
```
SimoneMaiwald: same as above

src/test/java/org/broadinstitute/dsm/DirectMethodTest.java
```diff
@@ -51,6 +56,42 @@ public void beforeTest() {
...
+    @Test
+    public void testGA(){
+        String realUa = "UA-12345678-1";// dummy value
+        GoogleAnalytics ga = GoogleAnalytics.builder()
+                .withConfig(new GoogleAnalyticsConfig().setGatherStats(true).setUserAgent("Custom User Agent"))
+                .withDefaultRequest(new DefaultRequest().userIp("127.0.0.1").trackingId(realUa))
+                .withTrackingId(realUa)
+                .build();
+        ga.resetStats();
+        EventHit eventHit = new EventHit("test-dsm", "action-test-dsm", "label-test-dsm", 1);
```
SimoneMaiwald, line 68: what are these strings? should they get moved somewhere else?

src/main/java/org/broadinstitute/dsm/analytics/GoogleAnalyticsMetricsTracker.java
```diff
@@ -0,0 +1,98 @@
...
+    private GoogleAnalyticsMetricsTracker() {
+        initStudyMetricTracker();
+    }
+
+    public static void setConfig(@NonNull Config config){
+        CONFIG = config;
+    }
+
+    public static GoogleAnalyticsMetricsTracker getInstance() {
+        if (instance == null) {
```
gmakharat: I recommend to initialize `new GoogleAnalyticsMetricsTracker();` directly on 16th line, with this approach Java will automatically provide lazy instantiation and thread safety, you won't have to use synchronized in this method anymore and volatile keyword on the field anymore, code will become much more cleaner as well.

src/main/java/org/broadinstitute/dsm/analytics/GoogleAnalyticsMetricsTracker.java
```diff
@@ -0,0 +1,98 @@
...
+                if (instance == null) {
+                    instance = new GoogleAnalyticsMetricsTracker();
+                }
+            }
+        }
+        return instance;
+    }
+
+    private GoogleAnalytics getMetricTracker() {
+        if (googleAnalyticsTrackers == null) {
```
gmakharat: I think you don't need to check whether `googleAnalyticsTrackers` is null or not, since this method is called by `sendAnalyticsMetrics`, the only way to use `sendAnalyticsMetrics` method is to get instance of `GoogleAnalyticsMetricsTracker`, to get instance of `GoogleAnalyticsMetricsTracker` client has to call `getInstance` which automatically means that `initStudyMetricTracker` will be called and at the point when client will call `sendAnalyticsMetrics` everything will be instantiated.

src/main/java/org/broadinstitute/dsm/analytics/GoogleAnalyticsMetricsTracker.java
```diff
@@ -0,0 +1,98 @@
...
+
+            initStudyMetricTracker();
+        }
+        return googleAnalyticsTrackers;
+    }
+
+    private synchronized void initStudyMetricTracker() {
```
gmakharat: if you change the code according the the comment about initializing `GoogleAnalyticsMetricsTracker` directly on field declaration, you will not need `synchronized` keyword anymore.

src/main/java/org/broadinstitute/dsm/model/filter/participant/BaseFilterParticipantList.java
```diff
@@ -106,9 +109,13 @@ public BaseFilterParticipantList(String jsonBody) {
             }
 
             logger.info("Found query conditions for " + mergeConditions.size() + " tables");
+            GoogleAnalyticsMetricsTracker.getInstance().sendAnalyticsMetrics(ddpInstance.getName(), GoogleAnalyticsMetrics.EVENT_CATEGORY_PARTICIPANT_LIST,
+                    GoogleAnalyticsMetrics.EVENT_ACTION_PARTICIPANT_LIST, GoogleAnalyticsMetrics.EVENT_LABEL_PARTICIPANT_LIST,  Math.toIntExact((System.currentTimeMillis() - timer)/1000));
```
gmakharat: extracting this code `Math.toIntExact((System.currentTimeMillis() - timer)/1000)` as method with meaningful name would be useful for quick understanding in the future.

PR #323, opened 2021-12-03
DDP-7000_TissueAssignee

pegahtah:


src/main/java/org/broadinstitute/dsm/model/TissueList.java
```diff
@@ -37,11 +37,13 @@
     private OncHistoryDetail oncHistoryDetails;
     private Tissue tissue;
     private String ddpParticipantId;
+    private String participantId;
 
-    public TissueList(OncHistoryDetail OncHistoryDetails, Tissue tissue, String ddpParticipantId) {
+    public TissueList(OncHistoryDetail OncHistoryDetails, Tissue tissue, String ddpParticipantId, String participantI) {
```
SimoneMaiwald, line 42: participantI should be participantId

PR #342, opened 2022-01-10
ElasticSearch feature | patchroute refactoring -> elastic export -> data migration -> refactoring participant list

ScalaChamp:
**Things have done:**

- Introduced OO principles to separate responsibilities into different classes.
- Created `BasePatch` abstract class as a parent of different classes, which are serving common feature.
- Introduced Template method design pattern in `BasePatch` to extract common code to avoid code duplication.
- Created abstract methods in `BasePatch` to specify what children should do to patch data.
- Introduced factory method to hide low level details about creation of special instance of `BasePatch`.

********************************************************************************************************************
- Introduced `Exportable` interface for having an abstraction for exporting data and mappings to ElasticSearch.
- Created `BaseExporter` (abstract class) and `ElasticExportAdapter` using "Adapter Design Pattern" to loose coupling from ElasticSearch API, which wraps all the necessary API-s from ElasticSearch library. 
- We created two DTO classes in order to reduce verbosity/complexity of `ElasticExportAdapter`, these classes are: `UpsertDataRequestPayload` and `UpsertMappingRequestPayload` which are used by `ElasticExportAdapter` in order to make request to ElasticSearch. 
- We introduced `Generator` and `Collector` abstractions in order to generate and collect data/mapping before making upsert request to ElasticSearch. 
- Concrete implementations of `Generator` and `Collector` are: `BaseGenerator` (abstract class), `MappingGenerator` and `SourceGenerator` which are responsible for generating and collecting either mapping or data. 
- Created another DTO class `GeneratorPayload` for `Generator` classes which are used generating mapping/data. 
- Introduced `Processor` abstraction to process data before "upserting" to ElasticSearch to avoid replacing existing ElasticSearch data. 
- Concrete implementations of `Processor` is: `CollectionProcessor` whose purpose is to work on data represented as "1 to many relationships". 
- Introduced `Parser` abstraction in order to parse text values to concrete types, it works for mapping and data as well.
- Concrete implementations of `Parser` are: `TypeParser` and `ValueParser` which are responsible for parsing types for mapping and parsing values for data. 
- Introduced `ExportFacade` class which uses "Facade Design Pattern" which is a front-facing protocol masking more underlying complexity and details of exporting to ElasticSearch. 
- Created another DTO class: `ExportFacadePayload` which is needed for transferring all the necessary data to `ExportFacade`
- Added `Util` class for ElasticSearch which is responsible for providing only utility methods which do not have side effects (any connection to database, network or 3rd party vendor). 
- Refactored legacy code based on new design.
- Eliminated the code whose functionality is covered by new feature. 

********************************************************************************************************************
- Made `BaseExporter` more generic by leaving only one method - `export()`. Created 2 subclasses: `ElasticDataExportAdapter` for exporting data and `ElasticMappingExportAdapter` for exporting mappings to ES. The subclasses use Adapter pattern in order to adapt to any occurring changes in ES API.
- Introduced the abstraction of `Merger` with method `merge` that helps us to merge different mappings into one final mapping, `Merger` can be used for other cases as well.
- Since it was agreed beforehand that we would use camel case pattern for keys we created a new method `underscoresToCamelCase()` in Util.java which handles this issue.
- Created an abstract class `BaseMigrator` which works for both cases (one to one, one to many). `BaseMigrator` uses following interfaces: `Generator` and `Exportable`.
- Added a new class `BulkExportFacade` which provides a simple API and hides the complexity associated with exporting data to ES.
- Introduced the class `BaseCollectionMigrator` which is a special case for classes which work on the data represented as `one-to-many` relationships.
- Created several concrete classes of `Migrator` abstraction: `KitRequestShippingMigrator`, `MedicalRecordMigrator`, `OncHistoryDetailsMigrator`, `ParticipantDataMigrator` and `ParticipantMigrator`.
- Used already existing `Pub/Sub` functionality as a control unit to make it easier to trigger exporting existing data to ES.
- Details of `Pub/Sub`: `taskType` = `ELASTIC_EXPORT`, `body`: `{"study": "<study_name>", "isMigration": true}`
- Made additional refactoring in order to clean up the code.

********************************************************************************************************************
- Created filtering for `ElasticSearch` based on existing filtering strategy.
- Introduced splitter abstraction for splitting filters by different criterias, e.g (<=, >=, <=>, LIKE, JSON_EXTRACT, DATE, STR_DATE and etc)
- Created the single and collection query builders to build `ElasticSearch` queries depending on operators.

P.S code was written using pair programming with @gmakharat .


src/main/java/org/broadinstitute/dsm/db/Participant.java
```diff
@@ -158,9 +186,10 @@ public static Participant getParticipant(@NonNull Map<String, Assignee> assignee
                 assigneeTissue = assignees.get(assigneeIdTissue).getName();
             }
         }
-        Participant participant = new Participant(rs.getString(DBConstants.PARTICIPANT_ID),
+        Participant participant = new Participant(rs.getLong(DBConstants.PARTICIPANT_ID),
                 rs.getString(DBConstants.DDP_PARTICIPANT_ID),
-                assigneeMR, assigneeTissue, realm,
+                rs.getLong(DBConstants.ASSIGNEE_ID_MR),
+                rs.getLong(DBConstants.ASSIGNEE_ID_TISSUE), realm,
```
SimoneMaiwald: the assigneeMR and assigneeTissue were names before and are now ids, can we change it back to names?

src/main/java/org/broadinstitute/dsm/db/dto/ddp/participant/ParticipantDataDto.java
```diff
@@ -1,29 +1,91 @@
...
+    @JsonGetter("fieldTypeId")
+    private String serializeFieldTypeId() {
+        return fieldTypeId;
+    }
+
+    @ColumnName(ParticipantDataDao.DATA)
+    @JsonProperty("dynamicFields")
     private String data;
+
+    @JsonProperty("dynamicFields")
```
SimoneMaiwald, line 68: does it make a problem if there are two JsonProperties with the same name?

src/main/java/org/broadinstitute/dsm/db/structure/DBElement.java
```diff
@@ -4,6 +4,7 @@
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
+import java.util.List;
```
SimoneMaiwald: unused import

src/main/java/org/broadinstitute/dsm/model/AbstractionWrapper.java
```diff
@@ -8,6 +8,7 @@
 import org.broadinstitute.dsm.db.AbstractionFinal;
 import org.broadinstitute.dsm.db.AbstractionGroup;
 import org.broadinstitute.dsm.db.structure.DBElement;
+import org.broadinstitute.dsm.model.patch.Patch;
```
SimoneMaiwald, line 22: unused import?

src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -194,7 +199,9 @@ else if (filter.isNotEmpty()) {
             }
         }
         else if (ADDITIONAL_VALUES.equals(filter.getType())) {
-            query = AND + " JSON_EXTRACT ( " + filter.getParentName() + DBConstants.ALIAS_DELIMITER + dbElement.getColumnName() + " , '$." + filter.getFilter2().getName() + "' ) ";
+            String jsonExtract = "JSON_EXTRACT ( ";
```
SimoneMaiwald: can we use the JSON_EXCTRACT string here as well. same as in L214

src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -297,6 +304,7 @@ private static String generateDateComparisonSql(Filter filter, DBElement dbEleme
             dateConverter = dbElement.getDateConverter();
             Instant instant = null;
             try {
+                // 29/07/2021
```
SimoneMaiwald, line 328: do we need that here?

src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -101,9 +106,9 @@ public static String getQueryStringForFiltering(@NonNull Filter filter, DBElemen
         String finalQuery = "";
         String query = "";
         String condition = "";
+        //simple is better than complex, KISS(Keep It Simple Stupid)
```
SimoneMaiwald, line 116: 😂

src/main/java/org/broadinstitute/dsm/model/elastic/export/parse/BaseParser.java
```diff
@@ -0,0 +1,88 @@
...
+
+    protected abstract Object forString(String value);
+
+    protected boolean isBoolean(String value) {
+        return convertBoolean(value).equalsIgnoreCase(Boolean.TRUE.toString()) ||
+               convertBoolean(value).equalsIgnoreCase(Boolean.FALSE.toString());
+    }
+
+    public String convertBoolean(String value) {
+        if ("'1'".equals(value) || "NOT'0'".equals(value) || "'true'".equals(value)) {
```
SimoneMaiwald, line 70: "NOT'0'" ? where is that

src/main/java/org/broadinstitute/dsm/model/elastic/filter/splitter/DateLowerSplitter.java
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.dsm.model.elastic.filter.splitter;
```
SimoneMaiwald, line 1: Looks like DateGreaterSplitter to me, could we not just use a DateSplitter and call it with the needed split parameter?

src/main/java/org/broadinstitute/dsm/db/Participant.java
SimoneMaiwald: @gmakharat @Anzori-Ghurtchumelia

src/main/java/org/broadinstitute/dsm/model/Filter.java
```diff
@@ -271,6 +259,48 @@ else if (filter.getFilter1() != null && filter.getFilter2().getValue() != null &
...
+                    int lessThanIndex = lessThan.indexOf(Filter.SMALLER_EQUALS_TRIMMED);
+                    moreThan = moreThan.substring(moreThanIndex);
+                    lessThan = lessThan.substring(lessThanIndex);
+                    query += moreThan + query + lessThan;
+                } else if (filter.isExactMatch()) {
+                    query += EQUALS + "'#'";
+                    query = query.replaceAll("#", String.valueOf(filter.getFilter1().getValue()));
+                } else {
+                    query += " " + LIKE + " '%#%'";
+                    query = query.replaceAll("#", String.valueOf(filter.getFilter1().getValue()));
```
SimoneMaiwald, line 290: possible sql injection?

src/main/java/org/broadinstitute/dsm/model/elastic/export/ElasticMappingExportAdapter.java
```diff
@@ -0,0 +1,26 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ElasticMappingExportAdapter extends BaseExporter {
+
+    private static final Logger logger = LoggerFactory.getLogger(ElasticMappingExportAdapter.class);
+
+    @Override
+    public void export() {
+        logger.info("initialize exporting mapping to ES");
```
SimoneMaiwald: I think logging the index would be nice here

src/main/java/org/broadinstitute/dsm/model/elastic/export/ExportFacade.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    private BaseGenerator.PropertyInfo getPropertyInfo() {
+        DBElement dbElement = PatchUtil.getColumnNameMap().get(exportFacadePayload.getFieldNameWithAlias());
+        return Util.TABLE_ALIAS_MAPPINGS.get(dbElement.getTableAlias());
+    }
+
+    private void upsertData(Map<String, Object> elasticDataToExport) {
+        RequestPayload requestPayload = new RequestPayload(exportFacadePayload.getIndex(), exportFacadePayload.getDocId());
+        logger.info("Built upsert data request payload");
```
SimoneMaiwald: maybe adding the index into the log as well

src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
```diff
@@ -0,0 +1,51 @@
...
+        for (Map.Entry<String, Object> entry: participantRecords.entrySet()) {
+            String participantId = entry.getKey();
+            participantId = getParticipantGuid(participantId, index);
+            if (StringUtils.isBlank(participantId)) continue;
+            Object participantDetails = entry.getValue();
+            transformObject(participantDetails);
+            Map<String, Object> finalMapToUpsert = generate();
+            bulkExportFacade.addDataToRequest(finalMapToUpsert, participantId);
+        }
+        logger.info("successfully filled bulk request with participants and their details");
```
SimoneMaiwald, line 39: please add index and study here as well

src/main/java/org/broadinstitute/dsm/model/elastic/migration/BaseMigrator.java
```diff
@@ -0,0 +1,51 @@
...
+
+    protected abstract void transformObject(Object object);
+
+    protected abstract Map<String, Object> getDataByRealm();
+
+    @Override
+    public void export() {
+        fillBulkRequestWithTransformedMap(getDataByRealm());
+        bulkExportFacade.executeBulkUpsert();
+        logger.info("finished migrating data to ES.");
```
SimoneMaiwald, line 50: or here, depending on where it is easier

src/main/java/org/broadinstitute/dsm/route/PatchRoute.java
```diff
@@ -68,501 +34,31 @@ public PatchRoute(@NonNull NotificationUtil notificationUtil, @NonNull PatchUtil
...
-                                throw new RuntimeException("DBElement not found in ColumnNameMap: " + patch.getNameValue().getName());
-                            }
-                            Patch.patch(String.valueOf(participantId), patch.getUser(), patch.getNameValue(), dbElement);
-                            return new Result(200, gson.toJson(Map.of(PARTICIPANT_ID, String.valueOf(participantId))));
-                        }
-                    }
-                }
-                throw new RuntimeException("Id and parentId was null");
+                Patch patch = GSON.fromJson(requestBody, Patch.class);
+                BasePatch patcher = PatchFactory.makePatch(patch, notificationUtil);
```
SimoneMaiwald, line 47: 😂

src/main/java/org/broadinstitute/dsm/util/KitUtil.java
```diff
@@ -141,7 +141,9 @@ public static void createLabel(List<KitRequestCreateLabel> kitsLabelTriggered) {
                         KitRequestShipping.updateRequest(kitLabelTriggered, ddpParticipant, kitLabelTriggered.getKitTyp(), kitLabelTriggered.getKitRequestSettings());
                     }
                     else {
-                        KitRequestShipping.deactivateKitRequest(kitLabelTriggered.getDsmKitRequestId(), "Participant not found", null, "System");
+                        KitRequestShipping.deactivateKitRequest(Long.parseLong(kitLabelTriggered.getDsmKitRequestId()), "Participant not found",
+                                null,
+                                "System");
```
SimoneMaiwald: I think we should change that to all caps "SYSTEM" the created_by for kit request is that way.

src/main/java/org/broadinstitute/dsm/db/dto/ddp/participant/ParticipantDataDto.java
gmakharat: No

src/main/java/org/broadinstitute/dsm/model/Filter.java
gmakharat: whole Filter class can have possible sql injection.

src/main/java/org/broadinstitute/dsm/model/elastic/export/parse/BaseParser.java
gmakharat: it comes from Filter class.

src/main/java/org/broadinstitute/dsm/model/elastic/filter/splitter/DateLowerSplitter.java
gmakharat: DateGreaterSplitter extends GreaterThanEqualsSplitter and DateLowerSplitter extends LessThanEqualsSplitter, one general class won't be able to extend two different class, but to avoid duplication of the same code we will add helper class for it.

src/main/java/org/broadinstitute/dsm/model/Filter.java
SimoneMaiwald: can you come up with an idea how to protect us from it?

gmakharat: our ES filtering is built depend on this class, if we would choose the way to eliminate this class fully we would have to change the whole logic of filtering, which would involve tons of changes on frontend side as well, that's why we decided to use already existing logic of frontend-backend for filtering, if we consider that this class is only used for filtering and we don't filter internal database anymore and we build ES queries according to this class, we can say that we won't have sql injection, but we need to keep in mind that we should try to avoid that class to filter something in database.

SimoneMaiwald: I am totally fine to keep that class and not change the logic, I was just wondering if we could add some checks on the value that would prevent sql injection…

gmakharat (approved):

PR #346, opened 2022-01-27
Added the Participant field to TissueList so we can filter on Assignees

pegahtah:


src/main/java/org/broadinstitute/dsm/model/TissueList.java
```diff
@@ -68,6 +80,9 @@ public TissueList(OncHistoryDetail OncHistoryDetails, Tissue tissue, String ddpP
                     OncHistoryDetail oncHistory = null;
                     while (rs.next()) {
                         String participantId = rs.getString(DBConstants.DDP_PARTICIPANT_ALIAS + DBConstants.ALIAS_DELIMITER + DBConstants.PARTICIPANT_ID);
+                        String ddpParticipantId = rs.getString(DBConstants.DDP_PARTICIPANT_ALIAS + DBConstants.ALIAS_DELIMITER + DBConstants.DDP_PARTICIPANT_ID);
+                        String assigneeIdTissue = rs.getString(DBConstants.DDP_PARTICIPANT_ALIAS + DBConstants.ALIAS_DELIMITER + DBConstants.ASSIGNEE_ID_TISSUE);
```
SimoneMaiwald, line 84: do we really want the id here or the name of the assignee?

pegahtah: id is better, that's what we store in the table and then the frontend already has function to convert name to id from the feature to assign tissue

SimoneMaiwald (approved):

